PR #5, opened 2024-07-14
Add support for TON

conorpp:


chain/ton/client.go
```diff
@@ -0,0 +1,512 @@
...
+	}
+
+	return sources, dests, nil
+}
+
+// Prioritize getting tx by msg-hash as it's deterministic offline.  Fallback to using chain-calculated tx hash.
+func (client *Client) FetchTonTxByHash(ctx context.Context, txHash xc.TxHash) (api.Transaction, error) {
+	transactions := &api.TransactionsData{}
+
+	err := client.get(fmt.Sprintf("api/v3/transactionsByMessage?direction=in&msg_hash=%s", url.QueryEscape(string(txHash))), transactions)
```
shevtsiv, line 356: Why do we limit it by `direction=in` here?

chain/ton/sign.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"crypto/ed25519"
+	"encoding/hex"
+	"fmt"
+
+	xc "github.com/cordialsys/crosschain"
+)
+
+// Signer for Solana
```
shevtsiv, line 11: These comments need some Solana=>TON renamings probably

chain/ton/tx/tx.go
```diff
@@ -0,0 +1,105 @@
...
+
+var _ xc.Tx = &Tx{}
+
+func (tx Tx) Hash() xc.TxHash {
+	if tx.ExternalMessage.Body == nil {
+		return ""
+	}
+	ext, err := tlb.ToCell(tx.ExternalMessage)
+	if err != nil {
+		panic(err)
```
shevtsiv, line 44: Just want to confirm we want to panic here and down below too.

chain/ton/builder.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	amountTlb, err := tlb.FromNano((*big.Int)(&amount), int(txBuilder.Asset.GetDecimals()))
+	if err != nil {
+		return nil, err
+	}
+
+	// Spend max 0.2 TON per Jetton transfer.  If we don't have 0.2 TON, we should
+	// lower the max to our balance less max-fees.
+	maxJettonFee := xc.NewAmountBlockchainFromUint64(200000000)
+	remainingTonBal := txInput.TonBalance.Sub(&txInput.EstimatedMaxFee)
```
shevtsiv, line 128: I'm not sure I fully understand the TON's mechanisms here, but I wonder if we should also account for the fees related to the excess amount return (https://docs.ton.org/develop/dapps/asset-processing/jettons#message-2-1). Or maybe it is already baked into the `txInput.EstimatedMaxFee`?

chain/ton/builder.go
```diff
@@ -0,0 +1,235 @@
...
+		return nil, fmt.Errorf("invalid TON token address %s: %v", txInput.TokenWallet, err)
+	}
+	amountTlb, err := tlb.FromNano((*big.Int)(&amount), int(txBuilder.Asset.GetDecimals()))
+	if err != nil {
+		return nil, err
+	}
+
+	// Spend max 0.2 TON per Jetton transfer.  If we don't have 0.2 TON, we should
+	// lower the max to our balance less max-fees.
+	maxJettonFee := xc.NewAmountBlockchainFromUint64(200000000)
```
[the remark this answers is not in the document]
conorpp: i'm not totally sure why TON requires some amount to be sent, in addition to all of the normal fees collected.  It seems to act like a max budget for the 3rd party token contract to take fees perhaps.  The excess gets sent back.

https://testnet.tonviewer.com/transaction/d464953f3bfd87058b0f3b531cbce11b77921387642e7eabccdd20e6ff6e80cb

I couldn't find a way to estimate a budget to use, so tried to set a high limit.  The Ton keeper wallet uses fixed 0.1 TON.

chain/ton/builder.go
```diff
@@ -0,0 +1,235 @@
...
+		ForwardPayload:      nil,
+	})
+	if err != nil {
+		return nil, err
+	}
+	return wallet.SimpleMessage(tokenWallet, maxFee, tokenBody), nil
+}
+
+func BuildV3UnsignedMessage(txInput *TxInput, messages []*wallet.Message) (*cell.Builder, error) {
+	if len(messages) > 4 {
```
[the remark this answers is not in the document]
conorpp: It's a limitation of TON, I had copied this from the ton lib.  will add a comment.

chain/ton/client.go
conorpp: Adding comment.  TON has concept of output transactions where the chain makes outputs in response to input transactions.  The "inputs" are the ones clients actually submitted.

PR #14, opened 2024-08-16
Add XRP to crosschain

mactus13:
- Add XRP address from pubkey functionality.


chain/xrp/client/client.go
```diff
@@ -0,0 +1,275 @@
...
+		return zero, fmt.Errorf("failed to marshal request payload: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", client.Url, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return zero, fmt.Errorf("failed to create new HTTP request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	httpClient := http.DefaultClient
```
conorpp, line 167: This should be done in the `Client` constructor.   Then would use `client.httpClient`.

chain/xrp/client/client_test.go
```diff
@@ -0,0 +1,161 @@
...
+	"fmt"
+	testtypes "github.com/cordialsys/crosschain/testutil/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"net/http"
+	"testing"
+
+	xc "github.com/cordialsys/crosschain"
+	xrpClient "github.com/cordialsys/crosschain/chain/xrp/client"
+	"github.com/cordialsys/crosschain/chain/xrp/tx"
```
conorpp, line 14: missing code?

chain/xrp/client/client.go
```diff
@@ -0,0 +1,275 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return zero, fmt.Errorf("failed to fetch balance, HTTP status: %s", resp.Status)
+	}
+
+	var response AccountInfoResponse
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return zero, fmt.Errorf("failed to decode response body: %w", err)
+	}
```
conorpp, line 181: This code will be duplicated in all functions.  I recommend making adding a helper function that implements http functionality for XRP endpoints.  E.g. something like:

```go
func send(httpClient *http.Client, method string, inputData interface{}, outputData interface{}) (error) {}
```

Check out existing clients elsewhere.

https://github.com/cordialsys/crosschain/blob/main/chain/evm/client/client_beacon.go#L107

https://github.com/cordialsys/crosschain/blob/main/client/services/kiln/send.go

chain/xrp/client/client.go
```diff
@@ -0,0 +1,275 @@
...
+	jsonPayload, err := json.Marshal(requestPayload)
+	if err != nil {
+		return zero, fmt.Errorf("failed to marshal request payload: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", client.Url, bytes.NewBuffer(jsonPayload))
+	if err != nil {
+		return zero, fmt.Errorf("failed to create new HTTP request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
conorpp, line 165: code duplication, should be in helper function

TODO
conorpp, line 1: debug file, delete

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,93 @@
...
+
+// NewNativeTransfer creates a new transfer for a native asset
+func (txBuilder TxBuilder) NewNativeTransfer(from xc.Address, to xc.Address, amount xc.AmountBlockchain, input xc.TxInput) (xc.Tx, error) {
+	txInput := input.(*TxInput)
+
+	signingPubKeyHex := hex.EncodeToString(txInput.Pubkey)
+	//txInput.XRPTx.SigningPubKey = signingPubKeyHex
+	fmt.Println(txInput)
+
+	jsonData, err := json.Marshal(txInput.XRPTx)
```
conorpp, line 46: You're only partially filling some fields in `txInput.XRPTx`, so it's hard to follow what is actually getting serialized/copied out.  It'd be more clear to explicitly define what our inputs are and then copy them here (I think just `LastLedgerSequence`, and `Sequence`)

chain/xrp/client/client.go
```diff
@@ -0,0 +1,580 @@
...
+}
+
+func (client *Client) FetchBaseInput(ctx context.Context, args xcbuilder.TransferArgs) (xrptxinput.TxInput, error) {
+	txInput := xrptxinput.NewTxInput()
+
+	XRPTransaction := txInput.XRPTx
+
+	XRPTransaction.Account = args.GetFrom()
+	XRPTransaction.Destination = args.GetTo()
+	XRPTransaction.Amount = args.GetAmount()
```
conorpp, line 230: These aren't really inputs from the ledger

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+// Serialize returns the serialized tx
+func (tx Tx) Serialize() ([]byte, error) {
+	if tx.XRPTx == nil {
+		return []byte{}, errors.New("missing XRP transaction")
+	}
+
+	//tx.XRPTx.SigningPubKey = hex.EncodeToString(txInput.Pubkey) // TODO: Possible this need to be added after signing it.
+	tx.XRPTx.SigningPubKey = *tx.SignPubKey
```
conorpp, line 99: i don't think it's correct to only add this here but not in the signature base.

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,93 @@
...
+		result["Sequence"] = int(sequence)
+	}
+
+	if flags, ok := result["Flags"].(float64); ok {
+		result["Flags"] = int(flags)
+	}
+
+	if lastLedgerSequence, ok := result["LastLedgerSequence"].(float64); ok {
+		result["LastLedgerSequence"] = int(lastLedgerSequence)
+	}
```
conorpp, line 67: if we defined our inputs explicitly (integer `Sequence` and `LastLedgerSequence`, then we wouldn't have to do this.

mactus13: I don't think it will work as:

"
 // To unmarshal JSON into an interface value,
// Unmarshal stores one of these in the interface value:
//
//   - bool, for JSON booleans
//   - float64, for JSON numbers
//   - string, for JSON strings
//   - []interface{}, for JSON arrays
//   - map[string]interface{}, for JSON objects
//   - nil for JSON null
//
"
.
Values are defined corect but the unmarshal operation changes them.

conorpp: There's no need to go through json unmarshaling.
1. Explicitly define the inputs required in the TxInput.

```go
type TxInput struct {
  // ...
   Sequence int
   LastLedgerSequence int
  // ...
}
```

2. Assign values to transaction object in TxBuilder.

```
result["Sequence"] = txInput.Sequence
result["LastLedgerSequence"] = txInput.LastLedgerSequence
```

No unmarshaling or conversion to floats needed.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,124 @@
...
+
+	return []xc.TxDataToSign{hash[:]}, nil
+
+}
+
+// AddSignatures adds a signature to Tx
+func (tx *Tx) AddSignatures(signatures ...xc.TxSignature) error {
+	for _, sig := range signatures {
+		signatureHex := hex.EncodeToString(sig)
+		tx.XRPTx.TxnSignature = signatureHex
```
conorpp, line 79: It looks like XRP uses DER encoding so we're need to convert it.  See example transaction, the binary byte length is not 64/65, but over 70 bytes.

https://xrpscan.com/tx/BE63DB29A62C16E02290562B28CFECA8870D65E8E1E0A605788643AAFE57C664

Bitcoin also uses DER so you can take example from that chain.

```go
	for i, rsvBytes := range signatures {
		r, s, err := DecodeEcdsaSignature(rsvBytes)
		if err != nil {
			return err
		}

		signature := ecdsa.NewSignature(&r, &s)
                tx.XRPTx.TxnSignature = signature.Serialize()
                // ...
    }
```

mactus13: Yes, you are right:
b'0D\x02 :\xc7\x0eC\x965\x0b\x1d\xb8l\x8b\xd1\x17~\xcf\x8fd6\xe1\x9e\xcf\xfa\xc5\xf3&M>\xae\x88|c\r\x02 )\xc1\x87x\x99\xa5\xab\xc6|=\xb4\xd3\x8f0\xa4\x98\x8d\x86\xcc\x96_O\x19\xf9i6\xad\xe6X\x85[\xa5'
.
This was obtained from Python library and it has 70bytes, length.

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,173 @@
...
+	result := make(map[string]interface{})
+	result["Account"] = string(xrpTx.Account)
+	result["Amount"] = xrpTx.Amount.StringValue
+	result["Destination"] = string(xrpTx.Destination)
+	result["Fee"] = xrpTx.Fee
+	result["Flags"] = int(xrpTx.Flags)
+	result["LastLedgerSequence"] = int(xrpTx.LastLedgerSequence)
+	result["Sequence"] = int(xrpTx.Sequence)
+	result["SigningPubKey"] = xrpTx.SigningPubKey
+	result["TransactionType"] = string(xrpTx.TransactionType)
```
conorpp, line 88: Please make a method to do this to remove duplicating it.

```go
func (txBuilder TxBuilder) renderToMap() map[string]interface{}
```

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,173 @@
...
+			Currency: tokenAsset,
+			Issuer:   tokenContract,
+			Value:    amount.String(),
+		},
+	}
+
+	amountResult := make(map[string]interface{})
+	amountResult["currency"] = tokenAsset
+	amountResult["issuer"] = tokenContract
+	amountResult["value"] = amount.String()
```
conorpp, line 133: Perhaps can build on `renderToMap ` example with function to override amount.

```go
func withTokenAmount(fields map[string]interface{}) map[string]interface{}

// ...

fields := txBuilder.renderToMap()
fields = withTokenAmount(fields)
```

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,228 @@
...
+	result["Account"] = string(xrpTx.Account)
+	result["Amount"] = XRPAmount.String()
+	result["Destination"] = string(xrpTx.Destination)
+	result["Fee"] = xrpTx.Fee
+	result["Flags"] = xrpTx.Flags
+	result["LastLedgerSequence"] = xrpTx.LastLedgerSequence
+	result["Sequence"] = xrpTx.Sequence
+	result["SigningPubKey"] = xrpTx.SigningPubKey
+	result["TransactionType"] = string(xrpTx.TransactionType)
+	result["TxnSignature"] = xrpTx.TxnSignature
```
conorpp, line 183: Duplicated.  I don't like how this rendering is done in both `tx` and `builder` modules.  I don't think it needs to be done in `builder` module, so i'd prefer to refactor to only do it in `tx` module.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,228 @@
...
+
+	firstHalf := hash[:32]
+	hashHex := hex.EncodeToString(firstHalf)
+	fmt.Println(hashHex)
+
+	return []byte(encodeTx), nil
+}
+
+// ExtractAssetAndContract parse assetContract and returns asset and contract
+func ExtractAssetAndContract(assetContract string) (string, string, error) {
```
conorpp, line 204: nit: change signature slightly to:

```go
func ExtractAssetAndContract(assetContract string) (asset string, contract string, err error) {
```

makes it slightly more clear

chain/xrp/tx_input/tx_input.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+func (input *TxInput) IndependentOf(other xc.TxInput) (independent bool) {
+	// are these two transactions independent (e.g. different sequences & utxos & expirations?)
+	if emvOther, ok := other.(*TxInput); ok {
+		return emvOther.Sequence != input.Sequence
+	}
+
+	return true
```
conorpp, line 85: Should be default false.  If we can't tell what the other input is, we can't know if there is independence.

chain/xrp/tx/tx_test.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+func TestTxHash(t *testing.T) {
+
+	tx1 := tx.Tx{}
+	hash := tx1.Hash()
+	require.Equal(t, "", string(hash))
+}
+
+func TestTxSighashes(t *testing.T) {
```
conorpp, line 19: I don't think this is a very good test.  it's basically testing it doesn't return an error.

I think it'd be much more useful to have test vectors with more realistic transactions and assert that their sighashes are expected values.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,228 @@
...
+		signatureBytes := signature.Serialize()
+		signatureHex := hex.EncodeToString(signatureBytes)
+		tx.XRPTx.TxnSignature = signatureHex
+		tx.TransactionSignature = append(tx.TransactionSignature, rsvBytes)
+	}
+
+	return nil
+}
+
+func DecodeEcdsaSignature(signature xc.TxSignature) (btcec.ModNScalar, btcec.ModNScalar, error) {
```
conorpp, line 121: I think this is duplicated from elsewhere in the codebase?  i.e. the bitcoin tx package.

Please just important the relevant package and re-use it instead.  You could make a subpackage in bitcoin tx.

chain/xrp/tx/tx_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+		encodeForSigningBytes, _ := hex.DecodeString(v.EncodeForSigningHex)
+		tx3 := tx.Tx{
+			XRPTx:            &tx.XRPTransaction{},
+			EncodeForSigning: encodeForSigningBytes,
+		}
+		sighashes, err := tx3.Sighashes()
+		require.Nil(t, err)
+		require.NotNil(t, sighashes)
+		require.Equal(t, sighashes[0], v.SigHash[0])
```
conorpp, line 109: nit; check length is 1

chain/xrp/tx/tx_test.go
```diff
@@ -0,0 +1,131 @@
...
+	require.Nil(t, sighashes2)
+
+	hexString := "5354580012000022000000002400070807201B000C44F96140000000014FB18068400000000000000A7321039543A0D3004CDA0904A09FB3710251C652D69EA338589279BC849D47A7B019A18114E2AFBD269D7DA5E2B9931CCBD62FAB5118A366188314BA4BEC4015A7CA2D99BE3319F488E0CA983D5506"
+	encodeForSigning, _ := hex.DecodeString(hexString)
+	tx3 := tx.Tx{
+		XRPTx:            &tx.XRPTransaction{},
+		EncodeForSigning: encodeForSigning,
+	}
+	sighashes3, err3 := tx3.Sighashes()
+	require.Nil(t, err3)
```
conorpp, line 40: please use `require.NoError(t, err3)`

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+	case *Amount:
+		fields["Amount"] = map[string]interface{}{
+			"currency": v.Currency,
+			"issuer":   v.Issuer,
+			"value":    v.Value,
+		}
+	case AmountBlockchain:
+		fields["Amount"] = map[string]interface{}{}
+	default:
+		fmt.Println("Invalid amount type")
```
conorpp, line 198: should return an error

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+	result["LastLedgerSequence"] = int(xrpTx.LastLedgerSequence)
+	result["Sequence"] = int(xrpTx.Sequence)
+	result["SigningPubKey"] = xrpTx.SigningPubKey
+	result["TransactionType"] = string(xrpTx.TransactionType)
+	result["TxnSignature"] = xrpTx.TxnSignature
+
+	return result
+}
+
+func WithTokenAmount(fields map[string]interface{}, amount interface{}) map[string]interface{} {
```
conorpp, line 185: `WithTokenAmount` not accurate since it's used for native XRP amount.  Perhaps a more typed way would be to define:

```go
func _setAnyAmount(fields map[string]interface{}, amount interface{}) (map[string]interface{}, error)

func SetXrpAmount(fields map[string]interface{}, amount string) (map[string]interface{}, error){
      return _setAnyAmount(fields, amount)
}

func SetTokenAmount(fields map[string]interface{}, amount *Amount) (map[string]interface{}, error) {
      return _setAnyAmount(fields, amount)
}

```

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+
+	encodeTxWithPrefix := TRANSACTION_HASH_PREFIX + string(encodeTx)
+
+	decodedBytes, err := hex.DecodeString(encodeTxWithPrefix)
+
+	hash := sha512.Sum512(decodedBytes)
+
+	firstHalf := hash[:32]
+	hashHex := hex.EncodeToString(firstHalf)
+	fmt.Println(hashHex)
```
conorpp, line 164: debugging.. please delete or refactor to:

```go
logrus.WithField("hash", HashFromTx(decodedBytes)).Debug("transaction hash")
```

Then we can opt in to see debugging logs by passing `-vv` flags, or the external application using the library can configure.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return []byte(encodeTx), nil
```
conorpp, line 166: I think it's confusing to return a "bytes-encoded-as-hex-string" as bytes.  How do we know it's hex, base64, or base58 representation?  Better to just return the bytes being represented, i.e. `decodedBytes` instead.   It's then up to the caller of `.Serialize()` to decide if they want to convert to hex, base64, etc, or not.

You'll probably need to update the client SubmitTx method to convert to hex appropriately.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+	StringValue string  `json:"-"`
+	AmountValue *Amount `json:"-"`
+	IsString    bool    `json:"-"`
+}
+
+type Amount struct {
+	Currency string `json:"currency"`
+	Issuer   string `json:"issuer"`
+	Value    string `json:"value"`
+}
```
conorpp, line 45: I would rewrite this as 

```go
// XRP amount is either string or Token Amount
type XRPOrTokenAmount struct {
	XRPAmount string  `json:"-"`
	TokenAmount *Amount `json:"-"`
}

type TokenAmount struct {
	Currency string `json:"currency"`
	Issuer   string `json:"issuer"`
	Value    string `json:"value"`
}
```

Don't need `.IsString` bool, can just check if `XRPAmount` is empty string or not.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+
+// Tx for Template
+type Tx struct {
+	XRPTx                *XRPTransaction
+	SignPubKey           []byte
+	EncodeForSigning     []byte
```
conorpp, line 51: I think better that we drop this and just compute it on demand for .SigHashes()

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return txInput, nil
+}
+
+// SubmitTx submits a Template tx
+func (client *Client) SubmitTx(ctx context.Context, txInput xc.Tx) error {
+	serializedTxInput, err := txInput.Serialize()
```
conorpp, line 446: convert to hex, shouldn't assume `txInput.Serialize()` returns hex.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+	}
+
+	confirmations := ledgerResponse.Result.LedgerCurrentIndex - txResponse.Result.Sequence
+
+	explorer := client.Asset.GetChain().ExplorerURL + "/tx/" + txResponse.Result.Hash + "?cluster=" + client.Asset.GetChain().Net
+
+	var sources []*xc.LegacyTxInfoEndpoint
+	sources = append(sources, &xc.LegacyTxInfoEndpoint{
+		Address: xc.Address(txResponse.Result.Account),
+		Amount:  xc.NewAmountBlockchainFromStr(txResponse.Result.Amount),
```
conorpp, line 521: contract address? for token amounts

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+	DeliveredAmount   string          `json:"delivered_amount,omitempty"`
+}
+
+type AffectedNodes struct {
+	CreatedNode  *CreatedNode  `json:"CreatedNode,omitempty"`
+	ModifiedNode *ModifiedNode `json:"ModifiedNode,omitempty"`
+}
+
+// UnmarshalJSON for AffectedNode
+func (an *AffectedNodes) UnmarshalJSON(data []byte) error {
```
conorpp, line 244: why the custom unmarshal?  if the default unmarshal doesn't work, would like to see comment to understand what's causing issues.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+		return nil
+	}
+
+	return fmt.Errorf("TakerGets is neither a string nor an Amount")
+}
+
+type TakerPays struct {
+	StringValue string  `json:"-"`
+	AmountValue *Amount `json:"-"`
+	IsString    bool    `json:"-"`
```
conorpp, line 209: drop IsString, can just check the string zero value.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+
+	var amount Amount
+	if err := json.Unmarshal(data, &amount); err == nil {
+		tg.AmountValue = &amount
+		tg.IsString = false
+		return nil
+	}
+
+	return fmt.Errorf("TakerGets is neither a string nor an Amount")
+}
```
conorpp, line 204: this is the same as TakerPays.  Perhaps delete and just rename to `TakeGetsOrPays`

chain/xrp/client/client.go
```diff
@@ -0,0 +1,718 @@
...
+	}, nil
+}
+
+const MethodPost string = "POST"
+
+const AccountInfo string = "account_info"
+const AccountLines string = "account_lines"
+const Transaction string = "tx"
+const Ledger string = "ledger"
+const Submit string = "submit"
```
conorpp, line 43: A slightly better way to do this is:

```go
type JsonRPCMethod string
const AccountInfo JsonRPCMethod = "account_info"
const AccountLines JsonRPCMethod = "account_lines"
const Transaction JsonRPCMethod = "tx"
const Ledger JsonRPCMethod = "ledger"
const Submit JsonRPCMethod = "submit"

type LedgerIndex string

const Validated LedgerIndex = "validated"
const Current LedgerIndex = "current"

type AccountInfoRequest struct {
	Method JsonRPCMethod                  `json:"method"`
	Params []AccountInfoParamEntry `json:"params"`
}
```

conorpp: I'm also ok not typing these out, they each only tend to be used once and are pretty local.

chain/xrp/tx/tx.go
```diff
@@ -0,0 +1,229 @@
...
+	hashHex := hex.EncodeToString(firstHalf)
+	fmt.Println(hashHex)
+
+	return []byte(encodeTx), nil
+}
+
+func RenderToMap(xrpTx XRPTransaction) map[string]interface{} {
+	result := make(map[string]interface{})
+	result["Account"] = string(xrpTx.Account)
+	result["Amount"] = xrpTx.Amount
```
conorpp, line 172: I notice you later call `WithTokenAmount` after `RenderToMap`?  I think it'd be better if `RenderToMap` did the rendering of Amount correctly.

chain/xrp/client/client.go
mactus13: I thought address is the source/destination address.

chain/xrp/client/client.go
mactus13: I'd say it is necessary as all those custom UnmarshalJSON functions I've created are targeting the case when the field on which they are implemented can have different data types depending on the value.
.
The default Go UnmarshalJSON won't work ( I've also tested it ) as it cannot handle multiple types for a single field automatically. Also it expects a field to have a specific type and if it encounters a type mismatch it will throw error.
.
A viable example is with AffectedNodes in here which can hold elements of both CreatedNode and ModifiedNode.

chain/xrp/client/client.go
mactus13: But there is also a field named ContractAddress which I think it is as the names implies. So Address might be the wallet address.

chain/xrp/client/client.go
conorpp: To me it looks like it implements the behavior that is default already.   I.e. `"CreatedNode"`  always maps to decode as `CreatedNode`, and `"ModifiedNode"` always maps to decode as `ModifiedNode`.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,816 @@
...
+
+				sources = append(sources, &xc.LegacyTxInfoEndpoint{
+					Address: xc.Address(txResponse.Result.TakerPays.TokenAmount.Issuer),
+					Amount:  amount.ToBlockchain(15),
+					Asset:   txResponse.Result.TakerPays.TokenAmount.Currency,
+				})
+
+				for _, node := range txResponse.Result.Meta.AffectedNodes {
+					if node.ModifiedNode != nil {
+						if node.ModifiedNode.LedgerEntryType == "RippleState" {
```
conorpp, line 582: you could simplify this a bit by defining a helper method on the type of `txResponse.Result.Meta`.

```go
func (m *Meta) GetAffectedNodesByType(entryType string) []*AffectedNode {
    nodes := []*AffectedNode{}
    for _, node := range txResponse.Result.Meta.AffectedNodes {
	    if node.ModifiedNode != nil {
		    if node.ModifiedNode.LedgerEntryType == entryType {
                        nodes = append(nodes, node)
                    }
            }
    }

    return nodes
}

// later
for _, node := range txResponse.Result.Meta.GetAffectedNodesByType("RippleState") {
```

chain/xrp/client/client.go
```diff
@@ -0,0 +1,816 @@
...
+		if txResponse.Result.TakerGets != nil {
+			if txResponse.Result.TakerGets.XRPAmount != "" {
+				sources = append(sources, &xc.LegacyTxInfoEndpoint{
+					Address: xc.Address(txResponse.Result.Account),
+					Amount:  xc.NewAmountBlockchainFromStr(txResponse.Result.TakerGets.XRPAmount),
+				})
+
+				for _, node := range txResponse.Result.Meta.AffectedNodes {
+					if node.ModifiedNode != nil {
+						if node.ModifiedNode.LedgerEntryType == "AccountRoot" {
```
conorpp, line 508: Perhaps simplify

```go
for _, node := range txResponse.Result.Meta.GetAffectedNodesByType("AccountRoot") {
```

factory/drivers/factory.go
```diff
@@ -58,8 +62,10 @@ func NewClient(cfg ITask, driver Driver) (xclient.FullClient, error) {
 		return tron.NewClient(cfg)
 	case DriverTon:
 		return ton.NewClient(cfg)
+	case DriverXrp:
+		return xrpclient.NewClient(cfg)
 	}
-	return nil, errors.New("no client defined for chain: " + string(cfg.ID()))
+	return nil, errors.New("no client defined for chains: " + string(cfg.ID()))
```
conorpp, line 68: i think this should remain singular, there is one chain in question

chain/xrp/client/client.go
```diff
@@ -0,0 +1,816 @@
...
+	if err != nil {
+		return xclient.TxInfo{}, err
+	}
+
+	// Remap to new tx
+	return xclient.TxInfoFromLegacy(client.Asset.GetChain().Chain, legacyTxInfo, xclient.Account), nil
+}
+
+// FetchLegacyTxInfo Returns transaction info - legacy/old endpoint
+func (client *Client) FetchLegacyTxInfo(ctx context.Context, txHash xc.TxHash) (xc.LegacyTxInfo, error) {
```
conorpp, line 448: The logic in here regarding parsing the metadata to figure out the transfer flows isn't super clear to me and gets a bit nested.  I'd appreciate some helper functions to simplify (I added an example), and also some comments that explain how it works, at least at a high level.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,906 @@
...
+
+// Appending destinations for TakerPays token transactions
+func appendDestinationForTakerPaysToken(txResponse TransactionResponse, destinations *[]*xc.LegacyTxInfoEndpoint, amount xc.AmountHumanReadable) {
+	*destinations = append(*destinations, &xc.LegacyTxInfoEndpoint{
+		Address:         xc.Address(txResponse.Result.Account),
+		ContractAddress: xc.ContractAddress(txResponse.Result.TakerPays.TokenAmount.Issuer + "-" + txResponse.Result.TakerPays.TokenAmount.Currency),
+		Amount:          amount.ToBlockchain(15),
+		Asset:           txResponse.Result.TakerPays.TokenAmount.Currency,
+	})
+}
```
conorpp, line 650: I the function is only appending to a list, it's kind of too simple and makes it harder to read.  I think It'd be easier to read to just leave these inlined.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,906 @@
...
+		addSourceDestinationForPayment(txResponse, &sources, &destinations)
+	} else {
+		if txResponse.Result.TakerGets != nil {
+			handleTakerGets(txResponse, &sources, &destinations)
+		}
+
+		if txResponse.Result.TakerPays != nil {
+			handleTakerPays(txResponse, &sources, &destinations)
+		}
+	}
```
conorpp, line 544: I believe somewhere this does the "hard work" of parsing and understanding the tx metadata directly.  I wonder, if we just always do that, and ignore `Destination`, `TakerGets`, `TakerPays`, would it still work?

conorpp: Shouldn't the tx metadata still be accurate in all cases?

chain/xrp/client/client.go
```diff
@@ -0,0 +1,906 @@
...
+}
+
+// Append sources for TakerGets native transactions
+func appendSourceForTakerGetsXRP(txResponse TransactionResponse, sources *[]*xc.LegacyTxInfoEndpoint) {
+	*sources = append(*sources, &xc.LegacyTxInfoEndpoint{
+		Address: xc.Address(txResponse.Result.Account),
+		Amount:  xc.NewAmountBlockchainFromStr(txResponse.Result.TakerGets.XRPAmount),
+	})
+}
```
conorpp, line 622: Didn't discover that this was inaccurate?  I.e. this is like marking the bid or ask price as the actually transaction amount, which is not accurate.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,906 @@
...
+
+// Append destinations for TakerPays native transactions
+func appendSourceForTakerPaysXRP(txResponse TransactionResponse, destinations *[]*xc.LegacyTxInfoEndpoint) {
+	*destinations = append(*destinations, &xc.LegacyTxInfoEndpoint{
+		Address: xc.Address(txResponse.Result.Account),
+		Amount:  xc.NewAmountBlockchainFromStr(txResponse.Result.TakerPays.XRPAmount),
+	})
+}
```
conorpp, line 630: Again, I fear this is inaccurate?  I thought we parsed the tx metadata directly to get accurate view of transfers.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,906 @@
...
+		if err == nil {
+			*sources = append(*sources, &xc.LegacyTxInfoEndpoint{
+				Address: xc.Address(node.ModifiedNode.FinalFields.Account),
+				Amount:  convertedTransactedAmount.ToBlockchain(decimals),
+			})
+
+			// Check to see if TakerPays amount is the same as transacted amount
+			if (*destinations)[len(*destinations)-1].Amount.Uint64() != convertedTransactedAmount.ToBlockchain(15).Uint64() {
+				(*destinations)[len(*destinations)-1].Amount = convertedTransactedAmount.ToBlockchain(15)
+			}
```
conorpp, line 687: why not just take the value from `node.ModifiedNode` directly?

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+				var (
+					contract     string
+					finalBalance float64
+					parseErr     error
+				)
+
+				if node.ModifiedNode.FinalFields.Balance.XRPAmount != "" {
+					contract = ""
+				} else {
+					finalBalance, parseErr = strconv.ParseFloat(node.ModifiedNode.FinalFields.Balance.TokenAmount.Value, 10)
```
conorpp, line 559: please use `xc.AmountHuman` to parse as a decimal.  Using native floating point isn't safe as there can be truncation and rounding errors.

A float64 can safely store 15-17 decimals, which in theory would be enough for XRP.  But on other chains, there are assets that use more decimals/precision.  Let's stay consistent and use a safe decimal type (`xc.AmountHuman` in our case).

https://en.wikipedia.org/wiki/Double-precision_floating-point_format

Also if we do any math with floats, we can get small errors which is bad for and financial reporting.

https://stackoverflow.com/questions/3730019/why-not-use-double-or-float-to-represent-currency

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+	txInfo.AddTransfer(tf)
+
+	txInfo.Fees = txInfo.CalculateFees()
+
+	return txInfo, nil
+}
+
+func extractModifiedNodeBalance(node *ModifiedNode) (xc.AmountBlockchain, error) {
+	var (
+		finalBalance, previousBalance float64
```
conorpp, line 787: use `xc.AmountHuman`, no floats

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+	var transactedAmountString string
+	if decimals == 0 {
+		transactedAmountString = fmt.Sprintf("%.6f", transactedAmount)
+	} else {
+		transactedAmountString = fmt.Sprintf("%.15f", transactedAmount)
+	}
+
+	transactedAmountString = strings.TrimRight(strings.TrimRight(transactedAmountString, "0"), ".")
+
+	convertedTransactedAmount, conversionErr := xc.NewAmountHumanReadableFromStr(transactedAmountString)
```
conorpp, line 823: so we are going from `string -> float -> string -> decimal`?  Let's do `string -> decimal`.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+						nil,
+					)
+				} else if transactionType == "source" {
+					tf.AddSource(
+						xc.Address(modifiedNode.FinalFields.HighLimit.Issuer),
+						xc.ContractAddress(contract),
+						convertedTransactedAmount,
+						nil,
+					)
+				}
```
conorpp, line 642: I think there is a bug here where if it's an XRP amount, then `transactionType == ""` and no transfer gets updated.

I think you could simplify this logic and also not need this intermediate `transactionType` enum.  If you still need it, you should use a boolean instead.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+				} else {
+					tf.AddDestination(
+						xc.Address(createdNode.NewFields.LowLimit.Issuer),
+						xc.ContractAddress(contract),
+						convertedTransactedAmount.ToBlockchain(decimals),
+						nil,
+					)
+				}
+
+			} else if createdNode.LedgerEntryType == "RippleState" {
```
conorpp, line 713: I think this also has the same bug.  If `transactionType == ""` like when XRP is transferred, then no transfer gets updated.  Please refactor to not need this logic, or use bool.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+				} else {
+					newBalanceString = fmt.Sprintf("%.15f", math.Abs(newBalance))
+				}
+
+				newBalanceString = strings.TrimRight(strings.TrimRight(newBalanceString, "0"), ".")
+
+				convertedTransactedAmount, conversionErr := xc.NewAmountHumanReadableFromStr(newBalanceString)
+				if conversionErr != nil {
+					return xclient.TxInfo{}, conversionErr
+				}
```
conorpp, line 694: floats are not good.  also you're converting from `string -> float -> string -> decimal`, you can do `string -> decimal`.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+						contract = node.ModifiedNode.FinalFields.LowLimit.Issuer
+					}
+				}
+
+				convertedTransactedAmount, conversionErr := extractModifiedNodeBalance(node.ModifiedNode)
+				if conversionErr != nil {
+					return xclient.TxInfo{}, conversionErr
+				}
+
+				//if modifiedNode.FinalFields.Account != txResponse.Result.Account {
```
conorpp, line 627: why don't we use this?

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+
+	if balance == "" {
+		return zero, fmt.Errorf("empty balance returned for account: %s", address)
+	}
+
+	humanReadbleBalance, err := xc.NewAmountHumanReadableFromStr(balance)
+	if err != nil {
+		return zero, fmt.Errorf("failed to parse balance for account: %s", address)
+	}
+	return humanReadbleBalance.ToBlockchain(15), nil
```
conorpp, line 920: please define a constant `TRUSTLINE_DECIMALS` and reference that in place of `15` literals everywhere.

chain/xrp/client/client.go
```diff
@@ -0,0 +1,992 @@
...
+						contract = node.CreatedNode.NewFields.HighLimit.Issuer
+					} else {
+						//transactionType = "destination"
+						contract = node.CreatedNode.NewFields.LowLimit.Issuer
+					}
+				}
+
+				var newBalanceString string
+				if decimals == 0 {
+					newBalanceString = fmt.Sprintf("%.6f", math.Abs(newBalance))
```
conorpp, line 684: in theory this works because XRP has 6 decimals, but it's scary close to running into rounding errors.

chain/xrp/client/client_test.go
```diff
@@ -0,0 +1,698 @@
...
+		}
+
+		client, _ := xrpClient.NewClient(&xc.ChainConfig{URL: server.URL})
+		txInfo, err := client.FetchTxInfo(context.Background(), xc.TxHash(vector.txHash))
+
+		if vector.err != "" {
+			require.Equal(t, xc.LegacyTxInfo{}, txInfo)
+			require.ErrorContains(t, err, vector.err)
+		} else {
+			if testNo == 0 {
```
conorpp, line 554: don't do this, should be able to just have the expected result in the test case.

mactus13: It won't work like that.
As I am passing txResponse as JSON string.
.
If you don't like testCase == 0 or 1. I can do two tests.

mactus13: Or to create another JSON string object with expected results but imo it won't look better.

chain/xrp/client/client.go
```diff
@@ -39,6 +36,9 @@ func NewClient(cfgI xc.ITask) (*Client, error) {
 	}, nil
 }
 
+const TRUSTLINE_DECIMALS = 15
+const XRP_NATIVE_DECIMALS = 0
```
conorpp, line 40: i thought XRP has 6 decimals?

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	tokenAsset, tokenContract, err := xrptx.ExtractAssetAndContract(assetContract)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse and extract asset and contract: %w", err)
+	}
+
+	tokenAmountValue, err := convertFromBlockchainToHumanFormat(amount.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse amount to float: %w", err)
```
conorpp, line 103: "failed to parse amount to float"

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,161 @@
...
+		DestinationTag:     destinationTag,
+	}
+
+	return &xrptx.Tx{
+		XRPTx:      &xrpTx,
+		SignPubKey: txInput.PublicKey,
+	}, nil
+}
+
+func convertFromBlockchainToHumanFormat(input string) (string, error) {
```
conorpp, line 142: delete this

chain/xrp/builder/builder.go
```diff
@@ -0,0 +1,161 @@
...
+	if assetContract == "" {
+		return nil, fmt.Errorf("asset does not have a contract")
+	}
+
+	tokenAsset, tokenContract, err := xrptx.ExtractAssetAndContract(assetContract)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse and extract asset and contract: %w", err)
+	}
+
+	tokenAmountValue, err := convertFromBlockchainToHumanFormat(amount.String())
```
conorpp, line 101: use `amount.ToHuman(15)`, avoid floating point

chain/xrp/client/client.go
```diff
@@ -254,10 +262,139 @@ type DeletedNode struct {
...
+		} else {
+			if mnw.node.NewFields.LowLimit == nil {
+				return "", fmt.Errorf("empty LowLimit in NewFields")
+			}
+
+			contract = xc.ContractAddress(mnw.node.NewFields.LowLimit.Issuer)
+		}
+	}
+
+	return contract, nil
```
conorpp, line 342: A contract should be something like `USD-ABCD...` for it to be useful in other endpoints.  Need to prepend the symbol.

chain/xrp/client/client.go
```diff
@@ -274,6 +411,147 @@ type NewFields struct {
...
+			return true, nil
+		}
+
+	} else if mnw.node.LedgerEntryType == "RippleState" {
+		finalBalanceHumanReadable, err := xc.NewAmountHumanReadableFromStr(mnw.node.FinalFields.Balance.TokenAmount.Value)
+		if err != nil {
+			return false, err
+		}
+
+		finalBalanceBlockchain := finalBalanceHumanReadable.ToBlockchain(6)
```
conorpp, line 534: may not matter since comparing to zero, but this is the wrong decimals i believe.

chain/xrp/client/client_test.go
conorpp: I don't understand.  In both cases "0" and "1", you're comparing only the resulting `client.TxInfo` type.  So in your test case you should have something like:

```go
{
 // ...
  expectedTxInfo: client.TxInfo{
   // expected fields
  }
}
```

Doesn't really matter how the input json is setup.

chain/xrp/client/client.go
conorpp: we should use the max precision (15).  because if it's a token then there's a good chance it will get truncated.

E.g. if the amount is `0.0000000000001554354`, and we use this with 6 decimals, then we will get `0` out.

conorpp: The underlying type for `xc.AmountHuman` is `decimal.Decimal` which has more methods we can use.  Perhaps let's just do this.

```go
isZero := decimal.Decimal(finalBalanceHumanReadable).IsZero()
```

chain/xrp/client/client_test.go
```diff
@@ -547,42 +628,29 @@ func TestFetchTxInfo(t *testing.T) {
...
-				require.Contains(t, txInfo.Transfers[0].To[0].Address, "chains/addresses/rzvAXDKJnPi8m25HjXYiXAjJnzc7LGTfw")
-				require.Contains(t, txInfo.Transfers[0].To[0].Asset, "chains/assets/rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq")
-				require.Contains(t, txInfo.Transfers[0].To[0].Contract, "rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq")
-				require.Equal(t, txInfo.Transfers[0].To[0].Balance.String(), "2624741712800000")
-				require.Contains(t, txInfo.Transfers[0].To[1].Address, "chains/addresses/rs9ineLqrCzeAGS1bxsrW8x2n3bRJYAh3Q")
-				require.Equal(t, txInfo.Transfers[0].To[1].Balance.String(), "4862466")
+			for idx, fromTransfer := range vector.expectedTxInfo.Transfers[0].From {
+				require.Equal(t, fromTransfer.Address, vector.expectedTxInfo.Transfers[0].From[idx].Address)
+				require.Equal(t, fromTransfer.Balance.String(), vector.expectedTxInfo.Transfers[0].From[idx].Balance.String())
+				if fromTransfer.Contract != "" {
```
conorpp, line 640: why the conditional?  these should always be equal.

chain/xrp/client/client_test.go
```diff
@@ -547,42 +628,29 @@ func TestFetchTxInfo(t *testing.T) {
...
-				require.Equal(t, txInfo.Fees[0].Balance.String(), "12")
+			for idx, toTransfer := range vector.expectedTxInfo.Transfers[0].To {
+				require.Equal(t, toTransfer.Address, vector.expectedTxInfo.Transfers[0].To[idx].Address)
+				require.Equal(t, toTransfer.Balance.String(), vector.expectedTxInfo.Transfers[0].To[idx].Balance.String())
+				if toTransfer.Contract != "" {
+					require.Equal(t, toTransfer.Contract, vector.expectedTxInfo.Transfers[0].To[idx].Contract)
+				}
 			}
 
+			require.Equal(t, txInfo.Fees[0].Balance.String(), vector.expectedTxInfo.Fees[0].Balance.String())
```
conorpp, line 653: why only comparing a subset of the first transfer? 

I would expect to do:

```go
require.Equal(t, vector.expectedTxInfo, txInfo)
```

This is simpler and also tests all of the fields (e.g. block, confirmations, etc).

Alternatively you could break it down a bit:

```go
   // compare block info 
   require.Equal(t, vector.expectedTxInfo.Block, txInfo.Block)
   
   // compare other fields
  // ...
   
   // compare all tfs
  for i := range txInfo.Transfers {
    require.Equal(t, vector.expectedTxInfo.Transfers[i], txInfo.Transfers[i])
  }
```

but in both cases we test everything.

conorpp (approved):

PR #29, opened 2024-11-05
Setup integration tests for crosschain

mactus13:


dagger/dagger/main.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+}
+
+func (m *Dagger) NodeService(image string) *dagger.Service {
+	return dag.Container().
+		From(image).
+		WithWorkdir("/srv").
+		WithNewFile("index.html", "Hello, world!").
+		WithExec([]string{"bash", "-c", "(python3 /faucet.py &) ; /root/hardhat-boilerplate/start_node.sh"}).
```
conorpp, line 41: this is specific to EVM, likely you can just re-use the existing `CMD` from the image.

dagger/main.go
```diff
@@ -1,19 +1,43 @@
...
+	cmd := exec.Command("sh", "-c", "cat /dev/urandom | head -c 32 | xxd -p -c100")
+
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+
+	hexString := strings.TrimSpace(string(output))
+	return hexString, nil
+}
```
conorpp, line 37: This is a bad way to generate random numbers.  Could do it natively in go instead.

However, we don't really need this to be random.  I suggest deleting and using a fixed random hex PRIVATE_KEY, like I had before.

dagger/dagger_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"flag"
+	"fmt"
+	"os/exec"
+	"testing"
+)
+
+var chain string
+
+func init() {
+	flag.StringVar(&chain, "chain", "", "Specify the blockchain chain to test against")
```
conorpp, line 14: i'm not sure we can pass arguments to the tests like this (maybe I'm wrong).  We could also use environment variables.

dagger/dagger_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+	if chain == "" {
+		t.Fatal("Chain argument is required. Use --chain flag to specify it.")
+	}
+
+	ctx := context.Background()
+
+	address, err := handler.AddressHandler(ctx)
+	if err != nil {
+		t.Fatalf("Error when retriving wallet address: %v", err)
```
conorpp, line 119:
```go
import (
"github.com/stretchr/testify/suite"
)


require.NoError(t, err)
```

ci/utils.go
```diff
@@ -0,0 +1,247 @@
...
+	cmd.Flags().String("rpc", rpc, "RPC endpoint")
+	cmd.Flags().String("toAddress", toAddress, "Transfer receiving address")
+	cmd.Flags().String("contract", contract, "Asset contract")
+	cmd.Flags().String("decimals", decimals, "Amount decimals")
+	cmd.Flags().String("memo", memo, "Memo")
+	cmd.Flags().Duration("timeout", timeout, "Timeout")
+
+	executeCommand(t, cmd, map[string]string{"toAddress": toAddress, "transferedAmount": transferedAmount, "chain": chain, "rpc": rpc})
+	return txTransferStr
+}
```
conorpp, line 174: Doesn't need CLI or cobra.

ci/utils.go
```diff
@@ -0,0 +1,247 @@
...
+func executeCommand(t *testing.T, cmd *cobra.Command, flags map[string]string) {
+	for key, value := range flags {
+		_ = cmd.Flags().Set(key, value)
+	}
+	err := cmd.Execute()
+	require.NoError(t, err, "Command execution should succeed")
+}
+
+func getTestTokensFromFaucet(address, amount string) error {
+	url := "http://node-service:10001/chains/ETH/assets/ETH"
```
conorpp, line 202: This should be parameterized.

cmd/xc/handler/handlers.go
```diff
@@ -0,0 +1,253 @@
...
+	"time"
+
+	xc "github.com/cordialsys/crosschain"
+	"github.com/cordialsys/crosschain/cmd/xc/setup"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+func BalanceHandler(cmd *cobra.Command, args []string) (string, error) {
```
conorpp, line 19: Should pass only the required arguments with types.

cmd/xc/handler/handlers.go
```diff
@@ -0,0 +1,253 @@
...
+	balance, err := cli.FetchBalance(context.Background(), address)
+	if err != nil {
+		return "", fmt.Errorf("could not fetch balance for address %s: %v", address, err)
+	}
+	fmt.Println(balance)
+
+	return balance.String(), nil
+}
+
+func TxInputHandler(cmd *cobra.Command, args []string) (string, error) {
```
conorpp, line 40: Should pass only the required arguments with types.

cmd/xc/handler/handlers.go
```diff
@@ -0,0 +1,253 @@
...
+	if err != nil {
+		return "", fmt.Errorf("could not fetch transaction inputs: %v", err)
+	}
+
+	bz, _ := json.MarshalIndent(input, "", "  ")
+
+	return string(bz), nil
+}
+
+func TxTransferHandler(cmd *cobra.Command, args []string) (string, error) {
```
conorpp, line 65: Should pass only the required arguments with types.

cmd/xc/handler/handlers.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		bz, _ := json.MarshalIndent(info, "", "  ")
+
+		return string(bz), nil
+	}
+
+	return "", fmt.Errorf("could not find transaction that we submitted by hash %s", tx.Hash())
+}
+
+func AddressHandler(cmd *cobra.Command) (xc.Address, error) {
```
conorpp, line 211: Should pass typed arguments

ci/utils.go
```diff
@@ -0,0 +1,247 @@
...
+		},
+	}
+
+	cmd.Flags().String("chain", chain, "Blockchain chain to use")
+	cmd.Flags().String("foundAmount", foundAmount, "Amount to fund wallet")
+	cmd.Flags().String("rpc", rpc, "RPC endpoint")
+	cmd.Flags().String("transferedAmount", transferedAmount, "Amount to be transfered")
+	cmd.Flags().String("toAddress", toAddress, "Transfer receiving address")
+
+	executeCommand(t, cmd, map[string]string{"chain": chain})
```
conorpp, line 109: pretty messy.. not clear why cobra or CLI args are needed.

dagger/main.go
```diff
@@ -37,6 +36,7 @@ func (m *Dagger) TestChain(ctx context.Context, chain string, image string, sour
 		WithDirectory("/crosschain", source).
 		WithServiceBinding("node-service", nodeService).
 		WithEnvVariable("PRIVATE_KEY", privateKey).
-		WithExec([]string{fmt.Sprintf("go test crosschain/dagger/dagger_test.go -chain=%s", chain)}).
+		WithExec([]string{"sh", "-c", fmt.Sprintf("cd crosschain/ci/ && go test -run TestBalance --chain %s --foundAmount %s --rpc http://node-service:10000", chain, foundAmount)}).
+		WithExec([]string{"sh", "-c", fmt.Sprintf("cd crosschain/ci/ && go test -run TestTransfer --toAddress %s --transferedAmount %s --chain %s --foundAmount %s --rpc http://node-service:10000", toAddress, transferedAmount, chain, foundAmount)}).
```
conorpp, line 40: helpful to change these so `sh` isn't used, will make the dagger trace more clear.

dagger/main.go
```diff
@@ -0,0 +1,42 @@
...
+		WithExposedPort(10001).
+		AsService()
+
+	privateKey := "22194a8955e9233aa2f0a0206c8ea861e5fa92a613ab5c7e236a11de3f4bc9ad"
+	return dag.Container().
+		From("us-docker.pkg.dev/cordialsys/crosschain/rockylinux").
+		WithDirectory("/crosschain", source).
+		WithServiceBinding("node-service", nodeService).
+		WithEnvVariable("PRIVATE_KEY", privateKey).
+		WithExec([]string{"sh", "-c", fmt.Sprintf("cd crosschain/ci/ && go test -run TestBalance --chain %s --foundAmount %s --rpc http://node-service:10000", chain, foundAmount)}).
```
conorpp, line 39: Regarding excluding these tests from running as unit tests.

Should add build tag `ci` to crosschain/ci/ test files.

```
//go:build ci
```

Then when running from dagger, we should run:

```bash
go test -tags ci <... other args>
```

dagger/main.go
```diff
@@ -0,0 +1,42 @@
...
+// basic module structure as you get started with Dagger.
+//
+// Two functions have been pre-created. You can modify, delete, or add to them,
+// as needed. They demonstrate usage of arguments and return types using simple
+// echo and grep commands. The functions can be called from the dagger CLI or
+// from one of the SDKs.
+//
+// The first line in this comment block is a short description line and the
+// rest is a long description with more detail on the module's purpose or usage,
+// if appropriate. All modules should have a short description.
```
conorpp, line 13: delete

ci/utils.go
```diff
@@ -3,174 +3,255 @@ package ci
...
+	}
+
+	setup.OverrideChainSettings(chainConfig, rpcArgs)
+
+	ctx := context.Background()
+	ctx = context.WithValue(ctx, xcFactoryKey, xcFactory)
+	ctx = context.WithValue(ctx, chainConfigKey, chainConfig)
+	ctx = context.WithValue(ctx, rpcKey, rpc)
+
+	return ctx, nil
```
conorpp, line 75: This is better, but I think we don't need to use this context.  We can just use concrete types.

ci/utils.go
```diff
@@ -3,174 +3,255 @@ package ci
...
 }
 
-type Address struct {
-	Balance   string `json:"balance"`
-	Address   string `json:"address"`
-	AddressID string `json:"address_id"`
+func getAddress(ctx context.Context) (string, error) {
```
conorpp, line 78: no need to use context, just pass the arguments that are needed and use types.

ci/utils.go
```diff
@@ -3,174 +3,255 @@ package ci
...
-	cmd.Flags().String("toAddress", toAddress, "Transfer receiving address")
-
-	executeCommand(t, cmd, map[string]string{"chain": chain})
-	return addressStr
+func buildFaucetURL(host string, chainConfig *xc.ChainConfig) (string, error) {
+	if host == "" || chainConfig == nil {
+		return "", fmt.Errorf("baseURL, chainConfig must be non-empty")
+	}
+
+	// TODO: It feels hackish. Better ideas?
```
conorpp, line 165: use `chainConfig.Chain`

ci/utils.go
```diff
@@ -3,174 +3,255 @@ package ci
...
+	_, err := rand.Read(randomBytes)
+	if err != nil {
+		return "", fmt.Errorf("Failed to generate random bytes: %v", err)
+	}
+
+	// Save wallet PRIVATE_KEY to reset it after genereting new wallet address.
+	walletPKey := os.Getenv("PRIVATE_KEY")
+
+	privateKey := hex.EncodeToString(randomBytes)
+	os.Setenv("PRIVATE_KEY", privateKey)
```
conorpp, line 245: I feel like we have three ways of passing arguments to functions:
* concrete type
* embedded in `Context`
* setting environment variables to be read out later

It would be best if this could be organized in a way where we only use concrete types.

We use the `Context` for the CLI plumbing - not applicable here.
We use environment variables to read inputs from user to a CLI program - not applicable here.

Unless there's a benefit to using `Context` or environment variables here that I'm missing, should stick to concrete types.

ci/utils.go
mactus13: I've used this approach as this code is used in each function and I thought about placing it in a single method will be easy to maintain.

ci/utils.go
```diff
@@ -48,42 +46,46 @@ func validateRequiredFlags(t *testing.T, value, errorMsg string) {
...
+
+// 	setup.OverrideChainSettings(chainConfig, rpcArgs)
+
+// 	ctx := context.Background()
+// 	ctx = context.WithValue(ctx, xcFactoryKey, xcFactory)
+// 	ctx = context.WithValue(ctx, chainConfigKey, chainConfig)
+// 	ctx = context.WithValue(ctx, rpcKey, rpc)
+
+// 	return ctx, nil
+// }
```
conorpp, line 74: can delete

cmd/xc/handler/handlers.go
```diff
@@ -0,0 +1,224 @@
...
+	"strconv"
+	"time"
+
+	xc "github.com/cordialsys/crosschain"
+	"github.com/cordialsys/crosschain/factory"
+
+	"github.com/sirupsen/logrus"
+)
+
+func BalanceHandler(contract string, xcFactory *factory.Factory, chainConfig *xc.ChainConfig, args []string) (string, error) {
```
conorpp, line 18: I think these handlers aren't really worth it.  It's a lot of fluff just to ultimately do `cli.FetchBalance(...)` which is what we want to test.

Rather than make new abstractions I'd prefer to just use crosschain methods directly in the tests.  Not only will this be easier to follow, but I think it'll make for less code as well.

For example, all of these "handlers" they repeat code, like `xcFactory.NewClient` and `xcFactory.MustAddress`.  You're also still passing vague things like `arg []string` and it's not clear what the concrete types are.

Also passing `args []string` is still problematic

ci/utils.go
```diff
@@ -0,0 +1,332 @@
...
+	rpc   string
+)
+
+type contextKey string
+
+const (
+	xcFactoryKey   contextKey = "xcFactory"
+	chainConfigKey contextKey = "chainConfig"
+	rpcKey         contextKey = "rpc"
+)
```
conorpp, line 36: don't need

dagger/dagger_test.go
```diff
@@ -0,0 +1,169 @@
...
+	Balance   string `json:"balance"`
+	Address   string `json:"address"`
+	AddressID string `json:"address_id"`
+}
+
+type Fee struct {
+	Asset    string `json:"asset"`
+	Contract string `json:"contract"`
+	Balance  string `json:"balance"`
+}
```
conorpp, line 52: these types are still duplicated.  already defined in crosschain.

dagger/dagger_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	assert.Equal(t, walletBalance, amount)
+
+	_, err = handler.TxInputHandler(ctx, "", "", args) // Do we want to check anything from tx input?
+	if err != nil {
+		t.Fatalf("Error when retriving tx input: %v", err)
+	}
+
+	toAddress := "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266"
```
conorpp, line 143: this should be derived from a public key (or new private key -> public key -> address).  otherwise it's only going to work on EVM.  It's not valid on Solana or cosmos.

ci/utils.go
```diff
@@ -0,0 +1,332 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	os.Setenv("PRIVATE_KEY", walletPKey)
+
+	return addressStr, nil
+}
+
+func setupContext(t *testing.T, chain string, rpc string) context.Context {
```
conorpp, line 252: not used.

cmd/xc/handler/handlers.go
mactus13: Reworked it:
- removed handlers and made specific methods for common parts used by both command and tests so we won't have code duplicates
- managed every repetition we have in the code in terms as you've pointed above
- fix passing vague things and only use concrete types

ci/utils.go
mactus13: Reworked it:
- only using concrete types/ normal arguments. no context embedded things or environment variables set up.
- removed all CLI plumbing

PR #32, opened 2024-11-21
[LIN-211] Add crosschain support for BASE and NOBL chains.

mactus13:


asset.go
```diff
@@ -62,6 +62,8 @@ const (
 	TRX    = NativeAsset("TRX")    // TRON
 	SEI    = NativeAsset("SEI")    // Sei
 	XRP    = NativeAsset("XRP")    // XRP
+	BASE   = NativeAsset("BASE")   // BASE
+	NOBL   = NativeAsset("NOBL")   // Noble Chain
```
conorpp, line 66: Let's use `NOBLE` instead.  NOBL is actually a different chain: https://coinmarketcap.com/currencies/nobleblocks/

factory/defaults/chains/testnet.yaml
```diff
@@ -298,3 +298,16 @@ chains:
...
+   chain: BASE
+   driver: evm
+   chain_name: Base Protocol
+   decimals: 18
+   explorer_url: https://explorer.baseprotocol.com
+  NOBL:
+   chain: NOBL
+   driver: cosmos
+   chain_name: Noble
+   chain_prefix: noble
```
conorpp, line 311: cosmos chain assets always have a contract address so we need to configure one.  noble doesn't seem to have a "noble" asset similar to other cosmos chains, (not that I can find here https://explorer.nodestake.org/noble/supply).  See other cosmos chains.

Instead we can use one of the tokens, like `uusdc`.  This means crosschain will use uusdc for pay for gas.  We may need to update later to support choosing between available assets to pay for gas later on.

```
    chain_coin: uusdc
```

factory/defaults/chains/mainnet.yaml
```diff
@@ -513,3 +513,16 @@ chains:
     decimals: 6
     indexer_type: ripple
     polling_period: 3m
+  BASE:
```
conorpp, line 516: We should add DTI identifiers for chains on mainnet.  Also need to update XRP.

Please add a comment at the top of this file saying we can get DTI from https://download.dtif.org/data.json.
This way we can remember where to look :)

XRP: L6GTZC9G4
BASE: 6M4WCXL3X
NOBLE: (not released yet, leave blank)

asset.go
mactus13: Make sense.

PR #39, opened 2024-12-10
Implement XLM driver

PawelBis:
Add chain driver implementation for Stellar.

This required some additions to ChainConfig:
- TransactionValidityTime - time for which the transaction is considered valid
- NetworkPassphrase - used to hash/sighash transactions. More info: https://developers.stellar.org/docs/learn/encyclopedia/network-configuration/network-passphrases


asset.go
```diff
@@ -416,6 +422,9 @@ type ChainConfig struct {
 	XDti             string `yaml:"dti,omitempty"`
 	XCoinGeckoId     string `yaml:"coingecko_id,omitempty"`
 	XCoinMarketCapId string `yaml:"coinmarketcap_id,omitempty"`
+
+	// Number of seconds for which the transaction will be considered valid after submission. Used only by Stellar so far
+	TransactionValidityTime int `yaml:"transaction_validity_time"`
```
conorpp, line 427: use `time.Duration`, and then update configuration to use a string.  I think we can do much longer than 30s, let's say 2 hours (`2h`).

conorpp: otherwise simple things like gas fees changing, or temporary network connectivity issues will force people to have to re-sign, which can be annoying.  Better to allow a generous amount of time for the caller to resubmit in a robust way.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+func (client *Client) GetAccountUrl(address string) string {
+	return fmt.Sprintf("%s/accounts/%s", client.Url, address)
+}
+
+func (client *Client) GetLedgerUrl(sequence uint64) string {
+	return fmt.Sprintf("%s/ledgers/%d", client.Url, sequence)
+}
+
+func (client *Client) GetLatestLedgerUrl() string {
+	return fmt.Sprintf("%s/ledgers?order=desc&limit=1")
```
conorpp, line 299: broken format string

chain/xlm/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to post, HTTP status: %s", resp.Status)
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return fmt.Errorf("failed to decode response body: %w", err)
+	}
```
conorpp, line 321: Is there a body for error responses?  E.g. often there's a `status` + `message`.  Would be helpful to catch those and include in the error.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+func (client *Client) Get(url string, response any) error {
+	resp, err := client.HttpClient.Get(url)
+	if err != nil {
+		return fmt.Errorf("failed to execute HTTP request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to get, HTTP status: %s", resp.Status)
+	}
```
conorpp, line 336: error body?

amount.go
```diff
@@ -120,6 +120,16 @@ func MultiplyByFloat(amount AmountBlockchain, multiplier float64) AmountBlockcha
 	return result
 }
 
+// NewAmountBlockchainFromUint64 creates a new AmountBlockchain from a uint64
+func NewAmountBlockchainFromInt64(i64 int64) AmountBlockchain {
+	abs := uint64(i64)
+	if i64 < 0 {
+		abs = uint64(-i64)
```
conorpp, line 127: I think this isn't really behavior we want.  If you make a negative amount, you would expect `AmountBlockchain` to also be negative.

If this isn't the behavior, I would change the function signature to be:

```go
func NewAmountBlockchainFromInt64(i64 int64) (amount AmountBlockchain, ok bool) {
```

If it's negative, return `0, false` -- indicating that it failed and the default value (0) is used.

chain/xlm/client/client.go
PawelBis: Fixed

chain/xlm/client/client.go
PawelBis: Fixed

chain/xlm/client/client.go
PawelBis: Fixed

amount.go
PawelBis: Fixed

factory/defaults/chains/mainnet.yaml
```diff
@@ -673,3 +673,13 @@ chains:
...
+  XLM:
+    chain: XLM
+    driver: xlm
+    chain_name: xlm
+    chain_gas_price_default: 0.00001
+    decimals: 7
+    transaction_validity_time: "2h"
+    # Stellar testnet passphrase.
+    # Passphrases can be found here: https://developers.stellar.org/docs/learn/encyclopedia/network-configuration/network-passphrases
+    network_passphrase: "Test SDF Network ; September 2015"
```
conorpp, line 685: Is this supposed to be a secret?  Also this is in the mainnet.yaml, not testnet.yaml

If this is a sensitive secret, should use `.auth` and `.auth_secret`, and perhaps pass by env variable.

If this is not sensitive (it's some well known phrase), I don't think `network_passphrase` is the best name.  It's overly fitted to XLM -- we should try to be generic to avoid from blowing up this chain-configuration too much. 

If it's not a secret, IMO it fits as a chain-id, could use existing `.chain-id-str`.

amount.go
```diff
@@ -120,6 +120,16 @@ func MultiplyByFloat(amount AmountBlockchain, multiplier float64) AmountBlockcha
...
 
+// NewAmountBlockchainFromUint64 creates a new AmountBlockchain from a uint64
+func NewAmountBlockchainFromInt64(i64 int64) (AmountBlockchain, bool) {
+	if i64 < 0 {
+		return AmountBlockchain{}, false
+	}
+
+	bigInt := new(big.Int).SetUint64(uint64(i64))
+	return AmountBlockchain(*bigInt), true
+}
```
conorpp, line 131:
```go
func NewAmountBlockchainFromInt64(i64 int64) (AmountBlockchain, bool) {
	if i64 < 0 {
		return NewAmountBlockchainFromUint64(0)
	}
	return NewAmountBlockchainFromUint64(uint64(i64))
}
```

factory/defaults/chains/mainnet.yaml
PawelBis: I am not sure why they decided to name it this way, but I completely agree. It’s not a secret - I will use existing chain-id-str

chain/xlm/client/client.go
```diff
@@ -0,0 +1,469 @@
...
+var _ xclient.FullClient = &Client{}
+var _ xclient.ClientWithDecimals = &Client{}
+
+func NewClient(cfgI xc.ITask) (*Client, error) {
+	cfg := cfgI.GetChain()
+	networkPassphrase := cfg.NetworkPassphrase
+	if networkPassphrase == "" {
+		networkPassphrase = TESTNET_NETWORK_PASSPHRASE
+		logrus.Warnf("Stellar configuration is missing network passphrase. Using default testnet passphrase: %s", networkPassphrase)
+	}
```
conorpp, line 46: use .auth / .auth_secret instead.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,469 @@
...
+	txInput.MinLedgerSequence = ledger.Sequence
+
+	defaultFee := config.ChainGasPriceDefault
+	defaultIFee := uint32(defaultFee * math.Pow10(int(config.Decimals)))
+	txInput.BaseFee = defaultIFee
+	txInput.MaxFee = defaultIFee
+
+	minFee := config.ChainMinGasPrice
+	minIFee := uint32(minFee * math.Pow10(int(config.Decimals)))
+	txInput.MinFee = minIFee
```
conorpp, line 89: Can you comment here an explanation on how XLM fees wrok?  It seems like the fee is defined statically here -- will this continue to work if XLM fees increase?

asset.go
```diff
@@ -419,6 +425,15 @@ type ChainConfig struct {
 	XDti             string `yaml:"dti,omitempty"`
 	XCoinGeckoId     string `yaml:"coingecko_id,omitempty"`
 	XCoinMarketCapId string `yaml:"coinmarketcap_id,omitempty"`
+
+	// Number of seconds for which the transaction will be considered valid after submission.
+	// Format: `time.Duration` string
+	// Used only by Stellar so far
+	TransactionValidityTime string `yaml:"transaction_validity_time"`
```
conorpp, line 432: use `time.Duration` instead of `string`.  It should deserialize correctly.

conorpp: perhaps rename `TransactionValidityTime` to `TransactionActiveTime`, trying to simplify slightly

factory/defaults/chains/mainnet.yaml
PawelBis: Fixed

amount.go
PawelBis: Fixed

chain/xlm/client/client.go
PawelBis: I used `chain-id-str` instead - network pass phrases are publicly available: https://developers.stellar.org/docs/learn/encyclopedia/network-configuration/network-passphrases

asset.go
PawelBis: Fixed

chain/xlm/builder/builder_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	txEnvelope := nt.(*Tx).TxEnvelope
+	source := common.MustMuxedAccountFromAddres(from)
+	require.Equal(t, txEnvelope.SourceAccount(), source)
+
+	destination := common.MustMuxedAccountFromAddres(to)
+	payment, ok := txEnvelope.Operations()[0].Body.GetPaymentOp()
+	require.NotZero(t, ok)
+	require.Equal(t, payment.Destination, destination)
+
```
conorpp, line 70: any token specific assertions you could add?  seems the same as with native transfer.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,471 @@
...
+
+	return txInfo, nil
+}
+
+func (client *Client) FetchNativeBalance(ctx context.Context, address xc.Address) (xc.AmountBlockchain, error) {
+	return client.FetchBalanceByAsset(address, true, xc.NativeAsset("XLM"))
+}
+
+// Fetch asset balance by asset code
+func (client *Client) FetchBalanceByAsset(address xc.Address, fetchNative bool, asset xc.NativeAsset) (xc.AmountBlockchain, error) {
```
conorpp, line 289: `asset xc.NativeAsset` is used incorrectly here.  Should use `xc.NativeAsset` when it's actually one of the chain constants, but here it seems to be either contract or `xc.NativeAsset`.

Perhaps switch to `assetId string`.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,471 @@
...
+	// Asset code is omitted for native currency
+	if asset == "XLM" {
+		asset = ""
+	}
+	for _, balance := range response.Balances {
+		if balance.AssetType == types.AssetTypeLiquidityPoolShares {
+			continue
+		}
+
+		if balance.AssetCode == string(asset) {
```
conorpp, line 305: Does this work for tokens?  If I understand the asset code is something like `USD`, but to uniquely identify a token, need both asset code and the issuer address.

Example:

https://stellarchain.io/assets/USD-GDUKMGUGDZQK6YHYA5Z6AY2G4XDSZPSZ3SW5UN3ARVMO6QSRDWP5YLEX

If you use asset-code/`USD` by itself, then you could end up reporting on the wrong asset.

conorpp: XRP has the same issue (needing asset-code + contract).  I recommend changing your `XlmContractDetails`:

```go
type XlmContractDetails struct {
	Asset  xc.NativeAsset
	Issuer xc.Address
}
```

To the following:

```
type TokenContract string
```

And then valid a instantiation of `TokenContract` would be `USD-GDUKMGUGDZQK6YHYA5Z6AY2G4XDSZPSZ3SW5UN3ARVMO6QSRDWP5YLEX`, which will be compatible with the API, and can re-use the unit tests you've written already.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,471 @@
...
+		}
+		return fmt.Errorf("failed to get, HTTP status: %s, error:\n%v", resp.Status, string(body))
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&response); err != nil {
+		return fmt.Errorf("failed to decode response body: %w", err)
+	}
+
+	return nil
+}
```
conorpp, line 398: It can be helpful to add support for decoding errors in the event of non successful http error code.  I think the API returns a standard error type in event of error, something like https://docs.cordialapis.com/docs/treasury/ia0q061ou7zxc-error?

Otherwise if there is a failure, the return error message will be pretty opaque.

chain/xlm/common/common.go
```diff
@@ -0,0 +1,88 @@
...
+		return account, fmt.Errorf("failed to convert address to xdr.MuxedAccount")
+	}
+
+	return account, nil
+}
+
+func MustMuxedAccountFromAddres(address xc.Address) xdr.MuxedAccount {
+	acc, err := MuxedAccountFromAddress(address)
+	if err != nil {
+		panic("failed to convert address to xdr.MuxedAccount")
```
conorpp, line 84: panic(err)

chain/xlm/tx_input/tx_input.go
```diff
@@ -0,0 +1,99 @@
...
+
+	// Multiply the BaseFee and check if it's valid
+	newFee := input.BaseFee * uint32(multiplier.BigInt().Uint64())
+	if newFee < input.MinFee {
+		return fmt.Errorf(
+			"calculated gas(%d) is lower than minimal allowed gas(%d)",
+			newFee,
+			input.MinFee,
+		)
+	}
```
conorpp, line 90: The user's intent is to adjust the gas fees, but they wouldn't want transactions to get stuck here.  I think it would be better to either:

1. Clamp the fee to the limit
2. Permit using a fee beyond the limit anyway -- running into RPC error later on.

Also, i don't think `MinFee` is actually set anywhere?  It's not in in the config currently as `chain_min_gas_price` -- so it would end up as `0` and cause this to always return an error.  Probably better to clamp or just delete this min fee behavior unless it's actually relevant for XLM.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,471 @@
...
+
+	ledger, err := client.FetchLatestLedgerInfo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch ledger info: %w", err)
+	}
+
+	txInput.Sequence = currentSequence + 1
+	txInput.MinLedgerSequence = ledger.Sequence
+
+	defaultFee := config.ChainGasPriceDefault
```
conorpp, line 80: Can we estimate this dynamically?  It will be too tedious to make config updates to keep up with XLM fee market.

chain/xlm/tx/tx.go
```diff
@@ -0,0 +1,225 @@
...
+
+	switch envelope.Type {
+	case xdr.EnvelopeTypeEnvelopeTypeTx:
+		hash, err = HashTransactionV1(envelope.V1.Tx, passphrase)
+	case xdr.EnvelopeTypeEnvelopeTypeTxV0:
+		err = errors.New("XLM Transaction type 0 is unsupported")
+	case xdr.EnvelopeTypeEnvelopeTypeTxFeeBump:
+		err = errors.New("XLM FeeBump transaction is unsupported")
+	default:
+		err = errors.New("Invalid transaction type")
```
conorpp, line 115: just do:

```go
default:
    return fmt.Errorf("invalid transaction type: %s", envelope.Type)
```

chain/xlm/tx/tx.go
```diff
@@ -0,0 +1,225 @@
...
+	return TimeBounds{0, int64(0), false}
+}
+
+// Operation represents the operation types of the Stellar network.
+type Operation interface {
+	BuildXDR() (xdr.Operation, error)
+	FromXDR(xdrOp xdr.Operation) error
+	Validate() error
+	GetSourceAccount() string
+}
```
conorpp, line 51: I don't see why this is needed?  Is it copied from elsewhere?

chain/xlm/tx/tx.go
```diff
@@ -0,0 +1,225 @@
...
+	xdrTimeBounds := xdr.TimeBounds{
+		MinTime: xdr.TimePoint(prec.TimeBounds.MinTime),
+		MaxTime: xdr.TimePoint(prec.TimeBounds.MaxTime),
+	}
+
+	xdrCond.Type = xdr.PreconditionTypePrecondTime
+	xdrCond.TimeBounds = &xdrTimeBounds
+
+	return xdrCond
+}
```
conorpp, line 72: This appears to try to implement the `Operation` interface partially.  Seems something could be simplified?

If there is an interface that is required to be implemented, it's nice to add a typeguard to communicate that.

```go
var _ Operation = Preconditions{}
```

conorpp: I would separate this `Preconditions` [+ `Operation` if it's needed] into a separate file, perhaps xlm.go.

chain/xlm/client/client_test.go
```diff
@@ -0,0 +1,657 @@
...
+	badClient, err = client.NewClient(badConfig)
+	require.Nil(t, badClient)
+	require.ErrorContains(t, err, "ChainMinGasPrice cannot be negative")
+
+	config := &xc.ChainConfig{
+		ChainIDStr: "ChainID",
+	}
+	client, err := client.NewClient(config)
+	require.NotNil(t, client)
+	require.NoError(t, err)
```
conorpp, line 52: Perhaps could test the default configurations to have all required fields set (chain-id-str)

```go

	configs := []*factory.Factory{
		factory.NewFactory(&factory.FactoryOptions{}),
		factory.NewNotMainnetsFactory(&factory.FactoryOptions{}),
	}
	for _, cfg := range configs {
		for _, asset := range cfg.GetAllAssets() {
			if chain, ok := asset.(*xc.ChainConfig); ok {
				if chain.Chain.Driver() == xc.DriverXlm  {
					// validate that chain_id_str is set
					require.NotEmpty(chain.ChainIDStr, "XLM driver requires that the network-passphrase mapped to chain_id_str")
				}
			}
		}
	}
```

PawelBis: Configurations are tested in `TestNewClient`

chain/xlm/tx/tx.go
PawelBis: Operation interface was leftover - I thought that we will need to support other operations types.

I moved the `TimeBounds` and `Preconditions` to `xlm.go`.

chain/xlm/tx/tx.go
PawelBis: It's a leftover - at first I thought that we will need to support more operations. This is `stellar-go` operation interface.

chain/xlm/client/client_test.go
conorpp: I mean to test that `mainnet.yaml` and `testnet.yaml` defaults are configured correctly for XLM.

PawelBis: That's a great idea. On it

chain/xlm/client/client.go
PawelBis: Good catch. I added issuer comparison

chain/xlm/client/client.go
PawelBis: I added a comment and changed the mechanism

chain/xlm/builder/builder_test.go
PawelBis: Good catch, I added an AssetCode/Issuer check

chain/xlm/client/client.go
PawelBis: Fixed

chain/xlm/common/common.go
PawelBis: Fixed

chain/xlm/tx_input/tx_input.go
```diff
@@ -0,0 +1,99 @@
...
+
+// SetGasFeePriority implements xc.TxInputGasFeeMultiplier.SetGasFeePriority
+func (input *TxInput) SetGasFeePriority(priority xc.GasFeePriority) error {
+	multiplier, err := priority.GetDefault()
+	if err != nil {
+		return err
+	}
+
+	// Multiply the BaseFee and check if it's valid
+	newFee := input.BaseFee * uint32(multiplier.BigInt().Uint64())
```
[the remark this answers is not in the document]
PawelBis: Fixed

chain/xlm/tx_input/tx_input.go
PawelBis: Fixed

chain/xlm/tx/tx.go
PawelBis: Fixed

chain/xlm/client/client.go
PawelBis: I updated the error handling for both `GET` and `POST` methods, let me know if it's a valid approach.

chain/xlm/client/client.go
```diff
@@ -0,0 +1,477 @@
...
+			return nil, fmt.Errorf("failed to create tx input, tx amount(%s) greater than balance(%s)", amount.String(), balance.String())
+		}
+		balance = balance.Sub(&amount)
+	}
+
+	// Stellar requires the MaxFee specification, which defines the maximum amount
+	// we are willing to spend on the transaction fee.
+	maxFee := config.ChainMaxGasPrice
+	maxIFee := uint64(maxFee * math.Pow10(int(config.Decimals)))
+	blockchainFee := xc.NewAmountBlockchainFromUint64(maxIFee)
```
conorpp, line 98: I would construct `xc.AmountHuman` from `config.ChainMaxGasPrice`, then use `maxFee.ToBlockchain(config.Decimals)`

chain/xlm/client/client.go
```diff
@@ -0,0 +1,477 @@
...
+
+	ledger, err := client.FetchLatestLedgerInfo()
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch ledger info: %w", err)
+	}
+
+	txInput.Sequence = currentSequence + 1
+	txInput.MinLedgerSequence = ledger.Sequence
+
+	balance, err := accountDetails.GetNativeBalance()
```
conorpp, line 79: perhaps call it `remainingBalance`

factory/defaults/chains/mainnet.yaml
```diff
@@ -673,3 +673,13 @@ chains:
         asset_id: ripple
       indexing_co:
         chain_id: ripple
+  XLM:
+    chain: XLM
+    driver: xlm
+    chain_name: xlm
+    decimals: 7
+    transaction_active_time: "2h"
+    chain_max_gas_price: 1
```
conorpp, line 682: leave comment this is in "XLM"/lumens, not an atomic amount

chain/xlm/client/client.go
PawelBis: Good catch, fixed

chain/xlm/client/client.go
PawelBis: Fixed

factory/defaults/chains/mainnet.yaml
PawelBis: Fixed

PR #47, opened 2025-01-09
Add IndexingCo chain identifiers for mainnet

shevtsiv:


asset.go
```diff
@@ -311,6 +311,9 @@ type External struct {
 		// Coingecko's asset ID, if relevant
 		AssetId string `yaml:"asset_id,omitempty"`
 	} `yaml:"coin_gecko,omitempty"`
+	IndexingCo struct {
+		ChainId string `yaml:"chain_id,omitempty"`
```
conorpp, line 315: leave a comment where to find so we don't forget

```go
// IndexingCo ID's can be found at https://jiti.indexing.co/networks
```

conorpp:
should add also:

- `bsc` for `BNB`

asset.go
shevtsiv: Fixed, thanks!

conorpp (approved):

PR #55, opened 2025-02-03
Add support for btc taproot transactions

conorpp:


chain/bitcoin/address/address.go
```diff
@@ -37,17 +50,50 @@ func NewAddressDecoder() *BtcAddressDecoder {
...
+	}
+
 	params, err := params.GetParams(asset.GetChain())
 	if err != nil {
 		return AddressBuilder{}, err
 	}
+
+	algorithm, is_set := opts.GetAlgorithmType()
+	// Check if provided algo
+	if is_set == true {
```
conorpp, line 66:
```go
	algorithm, ok := opts.GetAlgorithmType()
	// Check if provided algo
	if ok == true {
```

chain/bitcoin/client/blockbook/blockbook.go
```diff
@@ -111,9 +114,23 @@ func (client *BlockbookClient) UnspentOutputs(ctx context.Context, addr xc.Addre
...
+		publicKey, err := btcec.ParsePubKey(rawPublicKey)
+		if err != nil {
+			return nil, err
+		}
+		taprootKey := txscript.ComputeTaprootKeyNoScript(publicKey)
+		witnessProgram := schnorr.SerializePubKey(taprootKey)
+		btcAddr, err = btcutil.NewAddressTaproot(witnessProgram, client.Chaincfg)
+		if err != nil {
+			return nil, err
+		}
```
conorpp, line 128: Why is this done?  Can you comment some explaination?

chain/bitcoin/tx_input/tx_input.go
```diff
@@ -230,6 +231,7 @@ func NewOutputs[UTXO UtxoI](unspentOutputs []UTXO, addressScript []byte) []Outpu
 			Value:        xc.NewAmountBlockchainFromUint64(u.GetValue()),
 			PubKeyScript: addressScript,
 		}
+		log.Infof("Utoxo hash: %v", u.GetTxHash())
```
conorpp, line 234:
```go
	log.Debugf("Utxo hash: %v", u.GetTxHash())
```

factory/signer/signer.go
```diff
@@ -145,6 +148,24 @@ func (s *Signer) Sign(data xc.TxDataToSign) (xc.TxSignature, error) {
...
+	// TODO: Remove BTC specific parts
+	case xc.Schnorr:
+		privKey, _ := btcec.PrivKeyFromBytes(s.privateKey)
+
+		// We have to provide an empty tweak since we are spending/producingkey-path
+		// key-path only
+		privKeyTweak := txscript.TweakTaprootPrivKey(
+			*privKey,
+			[]byte{},
+		)
```
conorpp, line 160: We shouldn't do this tweak since it doesn't properly model a "stock" remote signer.

Instead, the "tweak" details you should be able to apply when preparing the sighashes.  See this example:
https://github.com/conorpp/rust-bitcoin/blob/taproot-example2/bitcoin/examples/sign-tx-taproot.rs#L143-L170

conorpp: I think the sighashes are indeed already computed correctly, using `TapSighash` prefix like in the example.  Is this tweak actually needed?

chain/bitcoin/client/blockbook/blockbook.go
PawelBis: I removed both this and priv key tweak

chain/bitcoin/tx_input/tx_input.go
PawelBis: Good catch

factory/signer/signer.go
PawelBis: Removed, not required as you suggested

chain/bitcoin/address/address.go
PawelBis: Fixed

PR #66, opened 2025-02-26
Implement Filecoin driver

PawelBis:


chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+	// A delegated address for user-defined foreign actors:
+	// For example f410 is and Ethereum compatibile address space
+	ProtocolDelegated    = 4
+	EncodeStd            = "abcdefghijklmnopqrstuvwxyz234567"
+	BlsPublicKeyBytesLen = 48
+	// MaxSubaddressLen is the maximum length of a delegated address's sub-address.
+	MaxSubaddressLen = 54
+	// ChecksumHashLength defines the hash length used for calculating address checksums.
+	ChecksumHashLength = 4
+)
```
conorpp, line 38: Are these the same between mainnet and testnet?

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+
+	algorithm, ok := opts.GetAlgorithmType()
+	// Check if default algorithm is overridden
+	if ok {
+		if algorithm != xc.K256Sha256 {
+			return AddressBuilder{}, fmt.Errorf("unsupported address type: %s", algorithm)
+		}
+	} else {
+		algorithm = asset.GetChain().Driver.SignatureAlgorithm()
+	}
```
conorpp, line 68:
```go
algorithm := asset.GetChain().Driver.SignatureAlgorithm()
if alg, ok := opts.GetAlgorithmType(); ok {
    algorithm = alg
}
if algorithm != xc.K256Sha256 {
	return AddressBuilder{}, fmt.Errorf("unsupported address type: %s", algorithm)
}
```

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+	buf[0] = addressType
+	copy(buf[1:], publicKeyBytes)
+
+	address := prefix + fmt.Sprintf("%d", addressType) + string(buf)
+	return xc.Address(address), errors.New("not implementer")
+}
+
+func (ab AddressBuilder) GetAddressFromPublicKey(publicKeyBytes []byte) (xc.Address, error) {
+	switch ab.alghoritm {
+	case "bls":
```
conorpp, line 131: to confirm, this is `BLS12-381`?

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+		return "t", nil
+	default:
+		return "", fmt.Errorf("invalid network: %s", ab.network)
+	}
+}
+
+func ProtocolIdToBytes(rawAddress string) ([]byte, error) {
+	if len(rawAddress) > 20 {
+		return nil, fmt.Errorf("invalid address length: %d. Expected: 22", len(rawAddress))
+	}
```
conorpp, line 170: length check doesn't match the message

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+	case '0':
+		protocol = ProtocolId
+	case '1':
+		protocol = ProtocolSecp256k1
+	case '2':
+		protocol = ProtocolActor
+	case '3':
+		protocol = ProtocolBls
+	case '4':
+		protocol = ProtocolDelegated
```
conorpp, line 230: default case?

chain/fil/address/address_test.go
```diff
@@ -0,0 +1,53 @@
...
+		31, 143, 68, 176, 160, 207, 246, 216, 107, 109,
+		107, 114, 189, 15, 87, 193, 90, 238, 233, 101,
+		199, 8, 164, 180, 50, 52, 85, 105, 45, 148,
+		162, 245, 142, 210, 41, 99, 147, 101, 77, 21,
+		141, 199, 25, 28, 166, 44, 199, 173, 46, 123,
+		151, 16, 196, 180, 179,
+	})
+	require.NoError(t, err)
+	require.Equal(t, xc.Address("t13uhmulxtag3qfohj7h2nmtco7e7u3t3nxjdzi7q"), address)
+}
```
conorpp, line 53: perhaps add simple test coverage of `AddressToBytes`, `ProtocolDelegatedToBytes`, `ProtocolIdToBytes`

chain/fil/tx/tx.go
```diff
@@ -0,0 +1,180 @@
...
+		tx.Message.Nonce,
+		append([]byte{0}, tx.Message.Value.Bytes()...),
+		tx.Message.GasLimit,
+		append([]byte{0}, tx.Message.GasFeeCap.Bytes()...),
+		append([]byte{0}, tx.Message.GasPremium.Bytes()...),
+		tx.Message.Method,
+		append([]byte{}, tx.Message.Params...),
+	}
+
+	return cbor.Marshal(i, cbor.CanonicalEncOptions())
```
conorpp, line 144: TIL filecoin uses CBOR, same as webauthn :)

chain/fil/builder/builder.go
```diff
@@ -0,0 +1,53 @@
...
+		Version:    0,
+		To:         to,
+		From:       from,
+		Nonce:      txInput.Nonce,
+		Value:      xc.NewAmountBlockchainFromUint64(amount.Int().Uint64()),
+		GasLimit:   txInput.GasLimit,
+		GasFeeCap:  txInput.GasFeeCap,
+		GasPremium: txInput.GasPremium,
+		Method:     0,
+		Params:     []byte{},
```
conorpp, line 47: Perhaps use a constructor that takes `tx_input` + `xcbuilder.TransferArgs` as a input

```go
func NewMessage(args xcbuilder.TransferArgs,  input tx_input.TxInput)
```

chain/fil/builder/builder.go
```diff
@@ -0,0 +1,53 @@
...
+func (txBuilder TxBuilder) Transfer(args xcbuilder.TransferArgs, input xc.TxInput) (xc.Tx, error) {
+	return txBuilder.NewTransfer(args.GetFrom(), args.GetTo(), args.GetAmount(), input)
+}
+
+// Old transfer interface
+func (txBuilder TxBuilder) NewTransfer(xcFrom xc.Address, xcTo xc.Address, amount xc.AmountBlockchain, input xc.TxInput) (xc.Tx, error) {
+	txInput := input.(*TxInput)
+	from := string(xcFrom)
+	to := string(xcTo)
+
```
conorpp, line 36: Can we sanity check that the gas fee?  Guard against malicious RPC node suggesting to spend $1M on fees.  Or does Filecoin already cap fee natively?

See similar guard here:
https://github.com/cordialsys/crosschain/blob/main/chain/cosmos/builder/builder.go#L64

Perhaps 100 FIL?

chain/fil/client/client.go
```diff
@@ -0,0 +1,326 @@
...
+
+	amount := xc.NewAmountBlockchainFromStr(*response.Result)
+	return amount, nil
+}
+
+func (client *Client) FetchDecimals(ctx context.Context, contract xc.ContractAddress) (int, error) {
+	return int(client.Asset.GetChain().GetDecimals()), nil
+}
+
+// TODO: Consider fetching tipset and populating the response with all tipset transactions
```
conorpp, line 280: Since the tipset seems to only be a few blocks, I think it's worth doing that.  I added "sub-blocks" to Block type for TON blockchain which has a similar model.

https://github.com/cordialsys/crosschain/blob/main/chain/ton/client.go#L631

This gets used by one of our jobs to seed various transaction hashes to test downloading.  So not critical, and not necessary to add test coverage for this endpoint.

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+	}
+
+	return AddressBuilder{
+		network:   asset.GetChain().Net,
+		alghoritm: algorithm,
+	}, nil
+}
+
+func (ab AddressBuilder) GetSecp256k1Address(publicKeyBytes []byte) (xc.Address, error) {
+	pk, err := secp.ParsePubKey(publicKeyBytes)
```
conorpp, line 77: Can you use btcec dependency instead?  I think it does the same thing.  Will allow us to remove this `secp256k1` dependency.

```go
import "github.com/btcsuite/btcd/btcec/v2"

	pubKey, err := btcec.ParsePubKey(serializedPubKey)
	if err != nil {
		return nil, err
	}
```

chain/fil/address/address.go
```diff
@@ -0,0 +1,365 @@
...
+	"errors"
+	"fmt"
+	"math"
+	"math/bits"
+	"strconv"
+	"strings"
+
+	xc "github.com/cordialsys/crosschain"
+	xcaddress "github.com/cordialsys/crosschain/address"
+	"github.com/dchest/blake2b"
```
conorpp, line 16: use `"golang.org/x/crypto/blake2b"` instead, then we can remove this dep

factory/drivers/factory.go
```diff
@@ -54,6 +58,8 @@ func NewClient(cfg ITask, driver Driver) (xclient.FullClient, error) {
 		return evmclient.NewClient(cfg)
 	case DriverEVMLegacy:
 		return evm_legacy.NewClient(cfg)
+	case DriverFil:
```
conorpp, line 61: I think it's a little more consistent to use `DriverFilecoin`

factory/defaults/chains/testnet.yaml
```diff
@@ -120,6 +120,12 @@ chains:
     external:
       indexing_co:
         chain_id: eth_holesky
+  FIL:
+    chain: FIL
+    driver: fil
+    chain_name: fil
```
conorpp, line 126: can we use (renaming driver to filecoin):

```
    driver: filecoin 
    chain_name: Filecoin
```

I think it's a little more consistent

chain/fil/address/address.go
PawelBis: Yes, only the prefix changes between testnet(`t`) and mainnet(`f`).

chain/fil/address/address.go
PawelBis: Yes, exactly

chain/fil/address/address.go
PawelBis: Good catch!

chain/fil/builder/builder.go
PawelBis: I added caps for both `GasLimit` and `GasFeeCap`.

Other than that - filecoin is **punishing** users that are abusing gas limit - part of the "overcharge" will be burned, and part will be returned.

chain/fil/address/address_test.go
PawelBis: Remove`ProtocolDelegatedToBytes` as it is not supported out of the box. Delegated addresses are for example EVM compatibile addresses, and this require some EVM compatibility layer.

chain/fil/client/client.go
PawelBis: Done. We can use combination of `ChainGetTipsetAfterHeight`, `ChainGetParentMessages` and `ChainGetBlock` methods for gathering necessary data.

chain/filecoin/client/client.go
```diff
@@ -0,0 +1,428 @@
...
+	if xcMaxGasPrice.IsZero() {
+		logger.Info("max gas price not set, using default: +s", DefaultMaxGasPrice.String())
+		xcMaxGasPrice = DefaultMaxGasPrice
+	}
+
+	gasLimit := cfg.ChainGasTip
+	if gasLimit == 0 {
+		logger.Info("chain gas tip not set, using default: +s", DefaultMaxGasPrice.String())
+		gasLimit = DefaultGasLimit
+	}
```
conorpp, line 58: This doesn't seem right.  GasLimit typically is not the same as GasTip.  

Gas is a unit-less concept, and it is what is being paid for.  A transaction could take 15M gas and this should be relatively constant, assuming the same type of transaction is being made.

Looking example on explorer explorer: https://filfox.info/en/message/bafy2bzacecv744s3qtyjzi2ttvyg5lu5caxmwvczcawppbr25n2lbj43jleeg?block=bafy2bzaceb7viav4sa4miannd6rbth57k7wx6xbzcslwknkxwne7zicd4advq&t=4

```
Gas Fee Cap: 157,884 attoFIL
Gas Premium: 156,704 attoFIL
Gas Limit: 28,556,847
Gas Used: 22,910,286
Base Fee: 107 attoFIL
```

Looks pretty similar to EVM.  The fee should be:

```python
(107 + 156704) * (10 ** -18) * 22910286 # 3.59 microFIL
```

The gasTip is used for incentive and would naturally map to "Gas Premium" (not gas limit).

The transaction could have taken more "gas" from the blockchain to execute, up to 28M.  This would have made
the fee:

```python
(107 + 156704) * (10 ** -18) * 28556847 # 4.47 microFIL
```

Also, the total gas price (base fee + incentive fee), can increase up to `157884`, the Gas Fee Cap.  While the GasPremium is fixed, the BaseFee is not.  The transaction could get stuck in the mempool and the base fee coul d increase.

So the absolute worst case for fee for this transaction would be:

```
(157884) * (10 ** -18) * 28556847 # 4.5 microFIL
```

PawelBis: This is Filecoin Gas walkthrough: https://spec.filecoin.io/systems/filecoin_vm/gas_fee/

In this case both `GasLimit` and `GasTip` refer to this unit-less concept. Basically - GasTip is what amount of gas is user allowing to be used (`GasLimit`) on Filecoin side. And `GasPremium` is basically a price per gas for miner.

So the fee (for your example) would be:
- BurnFee: `BaseFee * GasUsed`, `107 * 22910286 = 2451400602` - 2.4 nanoFil
- MinerFee: `GasLimit * GasPremium`, `28556847 * 156704 = 4474972152288` - 4474.9 nanoFil. `GasPremium` will be clamped to `GasFeeCap`.

There is also a concept of `Overburn` fee, but the amount is usually really low.

So according to those rules we are able to limit the fee using `GasLimit` and `GasFeeCap` values.

PawelBis: I adde a tests for `GasTip` and `ChainGasMaxFee` values and a short description of Filecoin fees above the `Client.FetchTxInput` method

PR #93, opened 2025-04-22
multi-transfer support for bitcoin chains

conorpp:


chain/bitcoin/builder.go
```diff
@@ -115,12 +116,90 @@ func (txBuilder TxBuilder) NewNativeTransfer(from xc.Address, to xc.Address, amo
 	tx := tx.Tx{
 		MsgTx: msgTx,
 
-		From:   from,
-		To:     to,
-		Amount: amount,
-		Input:  local_input,
+		// From:   from,
```
PawelBis, line 119: Leftover comment

chain/bitcoin/tx_input/tx_input.go
```diff
@@ -173,26 +158,29 @@ func FilterForMinUtxoSet(unspentOutputs []Output, targetAmount xc.AmountBlockcha
 	// 1. sort from lowest to higher
 	if len(unspentOutputs) > 1 {
 		sort.Slice(unspentOutputs, func(i, j int) bool {
-			return unspentOutputs[i].Value.Cmp(&unspentOutputs[j].Value) <= 0
+			return unspentOutputs[i].Value.Cmp(&unspentOutputs[j].Value) > 0
```
PawelBis, line 161: Is it intended? It's not in line with comment above anymore (it's descending order now)

chain/crosschain/multi.go
```diff
@@ -0,0 +1,36 @@
+package crosschain
+
+// import (
```
PawelBis, line 3: Leftover file?

cmd/xc/commands/address.go
```diff
@@ -0,0 +1,64 @@
+package commands
```
PawelBis, line 1: I love this split

chain/bitcoin/tx_input/tx_input.go
conorpp: Good catch, but yes i decided it's more clear and easier to explain this strategy.  

Before it took like the top N UTXO and then the bottom M UTXO.  This seems kind of arbitrary, so now it just takes the top (N+M) UTXO (where N is the minimum needed, and M is some extra amount for consolidation).

conorpp: Although now I see that the comment is out-of-date, as it sorts from high to low :)

PR #96, opened 2025-05-02
Add cardano support

PawelBis:
# Cardano
It supports multiple api options, however nothing is really native.

What I considered:
- https://github.com/cardano-foundation/cardano-rosetta - rosetta based api - transaction submission feels running in circles, and they require to submit transaction that is assembled on their side, which sounds dangerous
- https://developers.cardano.org/docs/get-started/ogmios - lack of fetching `utxo` from the account, they suggest running indexing node
- https://developers.cardano.org/docs/get-started/koios
- https://developers.cardano.org/docs/get-started/blockfrost/overview

In the end I decided to use `blockfrost` - it is a simple rest api, with really good documentation and great examples.

## Fees

Cardano uses "First come, first served" approach - there is no bidding. Fee calculation depends on two `protocol_parameters`: `min_fee_a` - cost per tx byte, and `min_fee_b` - fixed cost.
Fee formula: `size(tx) * min_fee_a + min_fee_b`.

https://docs.cardano.org/about-cardano/explore-more/fee-structure#:~:text=Cardano's%20fee%20structure%20is%20quite,a%2Fb%20are%20protocol%20parameters

## Tokens

Cardano supports tokens. Token UTXO are required to contain minimal value of UTXO, which is calculated from `tx without value` size, rounded up to bytes and multiplied by `coin_per_utxo_word` network parameter. More about this here: https://cardano-ledger.readthedocs.io/en/latest/explanations/min-utxo-mary.html and here: https://cardano-ledger.readthedocs.io/en/latest/explanations/min-utxo-alonzo.html

### DIFF
--- a/chain/cardano/client/client.go
+++ b/chain/cardano/client/client.go
@@ -74,7 +74,7 @@ func NewClient(cfgI xc.ITask) (*Client, error) {
 
 func (client *Client) FetchProtocolParameters(ctx context.Context) (types.ProtocolParameters, error) {
 	var protocolParameters types.ProtocolParameters
-	err := client.Request(ctx, GET, "/epochs/latest/parameters", "", &protocolParameters)
+	err := client.Request(ctx, GET, "/epochs/latest/parameters", nil, &protocolParameters)
 
 	return protocolParameters, err
 }
@@ -82,7 +82,7 @@ func (client *Client) FetchProtocolParameters(ctx context.Context) (types.Protoc
 func (client *Client) FetchUtxos(ctx context.Context, address xc.Address, contract xc.ContractAddress) ([]types.Utxo, error) {
 	path := fmt.Sprintf("/addresses/%s/utxos/%s?order=desc", string(address), string(contract))
 	var response []types.Utxo
-	err := client.Request(ctx, GET, path, "", &response)
+	err := client.Request(ctx, GET, path, nil, &response)
 	return response, err
 }
 
@@ -137,7 +137,7 @@ func (client *Client) FetchTransferInput(ctx context.Context, args xcbuilder.Tra
 	utxos = GetMinUtxoSet(utxos, targetAmount, contract)
 
 	var latestBlock types.Block
-	err = client.Request(ctx, GET, "/blocks/latest", "", &latestBlock)
+	err = client.Request(ctx, GET, "/blocks/latest", nil, &latestBlock)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch block info: %w", err)
 	}
@@ -173,10 +173,9 @@ func (client *Client) SubmitTx(ctx context.Context, tx xc.Tx) error {
 	if err != nil {
 		return fmt.Errorf("failed to serialize tx: %w", err)
 	}
-	cborHex := hex.EncodeToString(bytes)
 
 	var response string
-	err = client.Request(ctx, POST, "/tx/submit", cborHex, &response)
+	err = client.Request(ctx, POST, "/tx/submit", bytes, &response)
 	if err != nil {
 		return fmt.Errorf("failed to submit transaction: %w", err)
 	}
@@ -192,21 +191,21 @@ func (client *Client) FetchLegacyTxInfo(ctx context.Context, txHash xc.TxHash) (
 // Returns transaction info - new endpoint
 func (client *Client) FetchTxInfo(ctx context.Context, txHash xc.TxHash) (xclient.TxInfo, error) {
 	var latestBlock types.Block
-	err := client.Request(ctx, GET, "/blocks/latest", "", &latestBlock)
+	err := client.Request(ctx, GET, "/blocks/latest", nil, &latestBlock)
 	if err != nil {
 		return xclient.TxInfo{}, fmt.Errorf("failed to fetch latest block: %w", err)
 	}
 
 	var transactionInfo types.TransactionInfo
 	transactionPath := fmt.Sprintf("/txs/%s", string(txHash))
-	err = client.Request(ctx, GET, transactionPath, "", &transactionInfo)
+	err = client.Request(ctx, GET, transactionPath, nil, &transactionInfo)
 	if err != nil {
 		return xclient.TxInfo{}, fmt.Errorf("failed to fetch transaction info: %w", err)
 	}
 
 	var blockInfo types.Block
 	blockPath := fmt.Sprintf("/blocks/%d", transactionInfo.BlockHeight)
-	err = client.Request(ctx, GET, blockPath, "", &blockInfo)
+	err = client.Request(ctx, GET, blockPath, nil, &blockInfo)
 	if err != nil {
 		return xclient.TxInfo{}, fmt.Errorf("failed to fetch block info: %w", err)
 	}
@@ -230,7 +229,7 @@ func (client *Client) FetchTxInfo(ctx context.Context, txHash xc.TxHash) (xclien
 
 	var transactionUtxos types.TransactionUtxos
 	transactionUtxosPath := fmt.Sprintf("%s/utxos", transactionPath)
-	err = client.Request(ctx, GET, transactionUtxosPath, "", &transactionUtxos)
+	err = client.Request(ctx, GET, transactionUtxosPath, nil, &transactionUtxos)
 	if err != nil {
 		return xclient.TxInfo{}, fmt.Errorf("failed to fetch transaction utxos: %w", err)
 	}
@@ -276,7 +275,7 @@ func (client *Client) FetchTxInfo(ctx context.Context, txHash xc.TxHash) (xclien
 func (client *Client) FetchBalance(ctx context.Context, args *xclient.BalanceArgs) (xc.AmountBlockchain, error) {
 	path := fmt.Sprintf("/addresses/%s", string(args.Address()))
 	var getAddressInfoResponse types.GetAddressInfoResponse
-	err := client.Request(ctx, GET, path, "", &getAddressInfoResponse)
+	err := client.Request(ctx, GET, path, nil, &getAddressInfoResponse)
 	if err != nil {
 		return xc.AmountBlockchain{}, fmt.Errorf("failed to fetch address info: %w", err)
 	}
@@ -308,7 +307,7 @@ func (client *Client) FetchBlock(ctx context.Context, args *xclient.BlockArgs) (
 		blockPath = "/blocks/latest"
 	}
 	var block types.Block
-	err := client.Request(ctx, GET, blockPath, "", &block)
+	err := client.Request(ctx, GET, blockPath, nil, &block)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch block info: %w", err)
 	}
@@ -322,7 +321,7 @@ func (client *Client) FetchBlock(ctx context.Context, args *xclient.BlockArgs) (
 
 	blockTransactionsPaths := fmt.Sprintf("%s/txs", blockPath)
 	transactionHashes := make([]string, 0)
-	err = client.Request(ctx, GET, blockTransactionsPaths, "", &transactionHashes)
+	err = client.Request(ctx, GET, blockTransactionsPaths, nil, &transactionHashes)
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch block transactions: %w", err)
 	}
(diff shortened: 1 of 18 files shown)


chain/cardano/tx_input/tx_input.go
```diff
@@ -0,0 +1,88 @@
...
+func (input *TxInput) GetDriver() xc.Driver {
+	return xc.DriverCardano
+}
+
+// Cardano does not support fee bidding
+func (input *TxInput) SetGasFeePriority(other xc.GasFeePriority) error {
+	return nil
+}
+
+// FIXME: Cardano fee calculation requires a TX size
```
PawelBis, line 49: @conorpp I am not really sure how to approach this. Cardano doesn't support any FEE bidding and fee is strictly tx size dependent. Formula: `txSize * feePerByte + FixedFee` - I am not sure how to estimate it here.

Maybe we could consider moving `GetFeeLimit` to `tx.Tx` interface?

conorpp: Bitcoin works the same way.  Generally the transaction size is dependent on the number of UTXO being spent right?  You could assume an average size per UTXO.  The `FetchTransferInput` can output the estimated utxo size + fee-per-byte.  Then the total fee could be `len(utxo) * utxo_size * fee_per_byte`.

chain/cardano/tx/tx.go
```diff
@@ -0,0 +1,483 @@
...
+	"github.com/cordialsys/crosschain/chain/cardano/client/types"
+	"github.com/cordialsys/crosschain/chain/cardano/tx_input"
+	"github.com/cosmos/btcutil/bech32"
+	"github.com/fxamacker/cbor/v2"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/crypto/blake2b"
+)
+
+const (
+	// PolicyId is 28 byte hash of the token policy
```
conorpp, line 19: which hash?

chain/cardano/tx/tx.go
```diff
@@ -0,0 +1,483 @@
...
+	Signature []byte
+}
+
+var _ xc.Tx = &Tx{}
+
+// CalcMinUtxoValue calculates the minimum UTXO value for a given set of token amounts
+// based on the Cardano protocol parameters.
+// Base formula: https://cardano-ledger.readthedocs.io/en/latest/explanations/min-utxo-mary.html
+// Alonzo follow up: https://github.com/IntersectMBO/cardano-ledger/blob/master/doc/explanations/min-utxo-alonzo.rst
+func CalcMinUtxoValue(coinsPerUtxoWord xc.AmountBlockchain, policyHashToAmounts map[PolicyHash]TokenNameHexToAmount) xc.AmountBlockchain {
```
conorpp, line 208: This is the minimum amount of ADA that must be sent?  Is the token also itself technically a UTXO?

chain/cardano/tx/tx.go
```diff
@@ -0,0 +1,483 @@
...
+	// Make sure to zero out the dummy witness
+	defer func() {
+		tx.Witness.Keys = make([]*VKeyWitness, 0)
+	}()
+
+	// Serialize tx with dummy signature for fee estimation
+	txCbor, err := cbor.Marshal(tx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal tx: %w", err)
+	}
```
conorpp, line 405: This could be done in the client itself in FetchTransferInput, since it would have access to both `args xcbuilder.TransferArgs, input tx_input.TxInput`.  I recommend you refactor so that the Client makes it's own dummy cardano Tx for fee estimation (perhaps using 0 fee initially).

Then the real transaction builder can use the size-based estimation.

chain/cardano/tx/tx.go
PawelBis: Cardano utxo can contain multiple amounts. MinUtxoValue is a minimum value of ADA that should be placed alongside the token on the UTXO.

There is no way to send just `1 token`, without any ADA.

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	API_VERSION = "/api/v0"
+)
+
+// Client for Template
+type Client struct {
+	Asset               xc.ITask
+	Url                 string
+	Network             string
+	Logger              *log.Entry
+	BlockfrostProjectId string
```
conorpp, line 40: The project ID also counts as the authorzation?

factory/defaults/chains/mainnet.yaml
```diff
@@ -26,6 +26,15 @@ chains:
...
         asset_id: "6756"
+  ADA:
+    chain: ADA
+    driver: cardano
+    chain_id: cardano
+    network: mainnet
+    chain_name: Cardano
+    decimals: 6
+    fee_limit: "3.0"
+    auth: "raw:preprodkeXbfCzkmYDogwPw4gvM6fxvYG9W1Wcm"
```
conorpp, line 37: please delete or change to something like `env:BLOCKFROST_API_KEY`

factory/defaults/chains/testnet.yaml
```diff
@@ -2,6 +2,14 @@
...
 network: "testnet"
 chains:
+  ADA:
+    chain: ADA
+    driver: cardano
+    chain_id: cardano
+    network: preprod
+    chain_name: Cardano
+    decimals: 6
+    auth: "raw:preprodkeXbfCzkmYDogwPw4gvM6fxvYG9W1Wcm"
```
conorpp, line 12: please delete or change to something like `env:BLOCKFROST_API_KEY`

chain/cardano/client/client.go
PawelBis: Yes, blockfrost requires `project-id` header and it is generated per account, per network.

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+		return nil, fmt.Errorf("failed to fetch block info: %w", err)
+	}
+
+	protocolParams, err := client.FetchProtocolParameters(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch protocol parameters: %w", err)
+	}
+
+	baseFee := protocolParams.FixedFee
+	feePerByte := protocolParams.FeePerByte
```
conorpp, line 152: probably should instantiate a txBuilder + Tx to figure out the transaction size.

chain/cardano/tx/tx.go
PawelBis: I will rename it to `CalcMinAdaValue`

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	defer response.Body.Close()
+
+	buff, err := io.ReadAll(response.Body)
+	if err != nil {
+		return err
+	}
+
+	if response.StatusCode != http.StatusOK {
+		return fmt.Errorf("failed to send request, status: %s, error: %s", response.Status, buff)
+	}
```
conorpp, line 389: Is there an error type here or is it actually just returning text?

conorpp: It can be helpful to support the native error type (if there is one..) as we can use that to special case things like TransactionNotFound or TransactionAlreadyExists.

factory/defaults/chains/mainnet.yaml
```diff
@@ -26,6 +26,15 @@ chains:
       coin_market_cap:
         chain_id:
         asset_id: "6756"
+  ADA:
```
conorpp, line 29: add:
```
chain_coin: "lovelace"
```

This is to set the contract value for the native asset if there is one.  Help distinguish later on if we get in a situation like `chains/ADA/assets/ADA` vs `chains/ADA/assets/lovelace`.

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	xclient "github.com/cordialsys/crosschain/client"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	POST = "POST"
+	GET  = "GET"
+	// Cardano uses lovelace as the smallest unit of ada
+	// 1 lovelace = 0.000001 ada
+	Lovelace    = "lovelace"
```
conorpp, line 29: Can use `tx.Lovelace` or `chain.ChainCoin`.

factory/defaults/chains/mainnet.yaml
```diff
@@ -26,6 +26,15 @@ chains:
...
+    chain: ADA
+    driver: cardano
+    chain_id: cardano
+    network: mainnet
+    chain_name: Cardano
+    decimals: 6
+    fee_limit: "3.0"
```
conorpp, line 36: I'd set FeeLimit higher, perhaps `30.0` ADA, which is only like $20.  Treasury will record this trust-on-first-use style, and use as a sanity limit for future transactions.

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	}
+
+	utxos, err := client.FetchUtxos(ctx, args.GetFrom(), contract)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch utxos: %w", err)
+	}
+
+	// Include the fee in target amount
+	cfg := client.Asset.GetChain()
+	feeLimit := cfg.FeeLimit
```
conorpp, line 133: I'd map `GasBudgetDefault` instead.

* `GasBudgetDefault` is one of the chain _client_ configurations, which is easy to change.
* `FeeLimit` is chain _base_ configuration, meant for the user/custodial solution to use as a safety limit protect from RPC fee griefing.

chain/cardano/client/client.go
```diff
@@ -0,0 +1,407 @@
...
+	// Cardano uses lovelace as the smallest unit of ada
+	// 1 lovelace = 0.000001 ada
+	Lovelace    = "lovelace"
+	Ada         = "ADA"
+	API_VERSION = "/api/v0"
+)
+
+// Client for Template
+type Client struct {
+	Asset               xc.ITask
```
conorpp, line 36: To guard against using `FeeLimit` parameter mistake in client, perhaps change construction:

```go
type Client struct {
	Cfg               *xc.ChainClientConfig
        NativeAsset xc.NativeAsset
       // (other base options copied manually)
// ...
}
```

chain/cardano/client/client.go
PawelBis: Thanks! Implemented

factory/defaults/chains/mainnet.yaml
PawelBis: Lovelace is simply an unit returned by blockfrost api, it doesn't have contract address

chain/cardano/tx/tx.go
```diff
@@ -0,0 +1,516 @@
...
+
+const (
+	// PolicyId is 28 byte hash of the token policy
+	PolicyIdLen             = 56
+	FeeMargin               = 500
+	DefaultHeaderSize       = 6
+	PolicyIdSize            = 28
+	AssetNameOverhead       = 12
+	UtxoEntrySizeWithoutVal = 27
+	CoinsPerUtxoWord        = 37_037
```
conorpp, line 26: descriptions for these?  not clear why they are important

chain/cardano/tx/tx.go
```diff
@@ -233,7 +287,10 @@ func CalcMinUtxoValue(coinsPerUtxoWord xc.AmountBlockchain, policyHashToAmounts
 	sizeOfValue += DefaultHeaderSize
 
 	utxoEntrySize := xc.NewAmountBlockchainFromUint64(uint64(UtxoEntrySizeWithoutVal + sizeOfValue))
-	return utxoEntrySize.Mul(&coinsPerUtxoWord)
+	coinsPerUtxoWord := xc.NewAmountBlockchainFromUint64(uint64(CoinsPerUtxoWord))
+	utxoEntrySize = utxoEntrySize.Mul(&coinsPerUtxoWord)
+	adamount := xc.NewAmountBlockchainFromUint64(1500000)
+	return utxoEntrySize.Add(&adamount) // 1.5 ADA
```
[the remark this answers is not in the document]
PawelBis: Leftover for testing, good catch

chain/cardano/tx/tx.go
PawelBis: I added reference to calculation formula and parameter descriptions

conorpp (approved):

PR #137, opened 2025-07-23
Implement ICP driver

PawelBis:
# ICP Driver

ICP is a unique blockchain:
- Single transaction per block
- Each token has its own, distinct ledger
- No traditional `by hash` indexing - uses block indices instead
- Different addressing systems for ICP ledgers vs ICRC token ledgers

## API

ICP uses separate APIs:
- **ICP ledger** (canister: `ryjl3-tyaaa-aaaaa-aaaba-cai`) - uses legacy ICP API with AccountIdentifier addressing - [method list](https://dashboard.internetcomputer.org/canister/ryjl3-tyaaa-aaaaa-aaaba-cai)
- **Token ledgers** (e.g., ckBTC: `mxzaz-hqaaa-aaaar-qaada-cai`) - use ICRC-1/ICRC-3 APIs with Account-based addressing - [method list](https://dashboard.internetcomputer.org/canister/mxzaz-hqaaa-aaaar-qaada-cai)

The ICP ledger implements some ICRC-1/3 methods for compatibility, but due to the different addressing systems, we must support both:
- Legacy ICP addressing (AccountIdentifier) - used by most explorers
- ICRC addressing (Account = Principal + optional Subaccount) - used by token standards

## Indexing

ICP uses a unique indexing approach that differs from traditional blockchains. Instead of transaction hashes as primary identifiers, ICP uses **block indices** (e.g., block #2905248). This creates several challenges:

### Block Index Limitations
- Block indices are only assigned **after** successful transaction submission
- This complicates transaction lookup immediately after broadcasting
- Applications can't predict the block index before submission completes
- Ledgers use separate block indexing
- Impossible to fetch proper transaction, without correct `ledger` principal

### Workaround: Hash-Based Lookup
To address these limitations, we implemented a transaction-by-hash lookup mechanism:
1. Fetch index canister with `icrc160_get_index_principal`
2. Fetch the sender's transaction history from the index canister
3. Compute hashes for each transaction in their history
4. Match the computed hash against the target transaction hash
5. Retrieve full transaction details using the found block index

**Note:** This approach requires the **sender** parameter to limit the search scope, as scanning the entire ledger would be inefficient.


conorpp:
@PawelBis please fix CI

chain/internet_computer/builder/builder_test.go
```diff
@@ -0,0 +1,46 @@
+package builder_test
+
+import (
+	"testing"
+
+	xc "github.com/cordialsys/crosschain"
+	"github.com/cordialsys/crosschain/builder/buildertest"
+	"github.com/cordialsys/crosschain/chain/template/builder"
+	"github.com/cordialsys/crosschain/chain/template/tx_input"
```
conorpp, line 9: fix template

chain/internet_computer/candid/did/data.go
```diff
@@ -0,0 +1,231 @@
+package did
+
+import (
+	"fmt"
+	"math/big"
+	"strconv"
+	"strings"
+
+	"github.com/0x51-dev/upeg/parser"
```
conorpp, line 9: we should probably copy this into a sub-package to avoid the dependency.

chain/internet_computer/tx_input/tx_input.go
```diff
@@ -0,0 +1,68 @@
...
+	if !xc.IsTypeOf(other, input) {
+		return false
+	}
+
+	if input.IndependentOf(other) {
+		return false
+	}
+
+	return true
+}
```
conorpp, line 68: Is there any expiration for ICP transactions?  I think currently if a user retries a transaction, it will never work.  As the new attempt will use a new `CreatedAtTime` doesn't conflict at all with the previous transaction attempt.

cmd/xc/commands/transfer.go
```diff
@@ -67,9 +68,14 @@ func CmdTxTransfer() *cobra.Command {
 			}
 			algorithm, _ := cmd.Flags().GetString("algorithm")
 			addressArgs := []xcaddress.AddressOption{}
+			infoArgs := []xclient.GetTxInfoOption{}
 			if algorithm != "" {
 				addressArgs = append(addressArgs, xcaddress.OptionAlgorithm(xc.SignatureType(algorithm)))
 			}
+			if contract != "" {
+				addressArgs = append(addressArgs, xcaddress.OptionContract(xc.ContractAddress(contract)))
```
conorpp, line 76: Why is this done?  IMO it's not realistic for treasury to do the same.

People generate the address once, and then it's fixed.  It should be on the `TxBuilder` to use a token address format if needed.

address/args.go
```diff
@@ -6,10 +6,12 @@ import (
 
 type addressOptions struct {
 	algorithm *xc.SignatureType
+	contract  *xc.ContractAddress
 }
 
 type AddressOptions interface {
 	GetAlgorithmType() (xc.SignatureType, bool)
+	GetContract() (xc.ContractAddress, bool)
```
conorpp, line 14: I think we should remove this, as it's not reasonable for Treasury product to generate addresses with a token address argument.  It should be refactored so that `TxBuilder` will format token addresses as needed.  The public key should always be passed as it's a signing address (public key is always known).

chain/internet_computer/tx/tx.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+	} else {
+		toAccount, err := types.DecodeICRC1Account(string(args.GetTo()))
+		if err != nil {
+			return Tx{}, fmt.Errorf("failed to decode destination icrc1 address: %w", err)
+		}
+		fromAccount, err := types.DecodeICRC1Account(string(args.GetFrom()))
+		if err != nil {
+			return Tx{}, fmt.Errorf("failed to decode source icrc1 address: %w", err)
+		}
```
conorpp, line 81: should expect owner address here.  should instead derive the token address.

chain/internet_computer/client/client.go
```diff
@@ -0,0 +1,365 @@
...
+	args, _ := xcbuilder.NewTransferArgs(from, to, xc.NewAmountBlockchainFromUint64(1))
+	return client.FetchTransferInput(ctx, args)
+}
+
+// SubmitTx submits a InternetComputerProtocol tx
+func (client *Client) SubmitTx(ctx context.Context, tx xc.Tx) error {
+	icpTx, ok := tx.(*icptx.Tx)
+	if !ok {
+		return errors.New("invalid transaction type")
+	}
```
conorpp, line 135: This doesn't really work when deployed as an API.  As SubmitTx() will always get a binary serialized tx:

https://github.com/cordialsys/crosschain/blob/main/chain/crosschain/types/api.go#L188-L217

Should really only rely on the `tx.Serialize` output.

If we really need these distinct fields, I think we'd have to extend the xc.Tx interface a bit.  I'll propose something.

PawelBis: Thanks! Merged

chain/internet_computer/tx_input/tx_input.go
conorpp: Yeah, it would be better for us to use nonces in this case.

I wonder if we treat `created_at_time` as sort of nonce.  The client can set `created_at_time` to be the floor of the most recent 6th hour + number of transactions as seconds.  This would at least give a valid time period where retries could be done effectively.  But I think this might be too complicated, maybe we can revisit when ICP releases support for nonces officially.

chain/internet_computer/tx/tx.go
PawelBis: `args.GetTo()` should be a valid address, so we are covered

chain/internet_computer/tx_input/tx_input.go
PawelBis: Great catch! Instead of checking `currentInput.CreatedAtTime - old.CreatedAtTime > 24h` I simply check if `now - oldInput.CreatedAtTime > 24h` - it is more permissive, and still works correctly.

PR #164, opened 2025-10-07
SUI: Implement staking stake, unstake and balance

PawelBis:


chain/sui/builder_staking.go
```diff
@@ -0,0 +1,257 @@
...
+		Field1: []bcs.Argument{
+			// the last input has the amount
+			ArgumentInput(uint16(len(cmd_inputs))),
+		},
+	})
+	cmd_inputs = append(cmd_inputs, U64ToPure(amount.Uint64()))
+
+	// add system state input
+	systemStateArg := ArgumentInput(uint16(len(cmd_inputs)))
+	systemState, err := HexToObjectID("0x0000000000000000000000000000000000000000000000000000000000000005")
```
conorpp, line 61: Does this ever change?  Can we look it up as part of the staking input?

chain/sui/builder_staking.go
```diff
@@ -0,0 +1,257 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create transaction base: %w", err)
+	}
+
+	commands := make([]bcs.Command, 0)
+	cmd_inputs := make([]bcs.CallArg, 0)
+
+	// prepare system state input - it will be reused 'request_withdraw_stake' commands
+	systemStateInput := ArgumentInput(uint16(len(cmd_inputs)))
+	systemState, err := HexToObjectID("0x0000000000000000000000000000000000000000000000000000000000000005")
```
conorpp, line 140: refactor to declared variable

chain/sui/builder_staking.go
```diff
@@ -0,0 +1,257 @@
...
+		return nil, fmt.Errorf("failed to encode sui system state: %w", err)
+	}
+	cmd_inputs = append(cmd_inputs, &bcs.CallArg__Object{
+		Value: &bcs.ObjectArg__SharedObject{
+			Id:                   systemState,
+			InitialSharedVersion: 1,
+			Mutable:              true,
+		},
+	})
+	packageId, err := HexToObjectID("0x0000000000000000000000000000000000000000000000000000000000000003")
```
conorpp, line 151: refactor to common variable

chain/sui/builder_staking.go
```diff
@@ -0,0 +1,257 @@
...
+					systemStateInput,
+					stakeIdInput,
+				},
+			},
+		})
+	}
+
+	// check if we have to prepare a split operation
+	if unstakeInput.StakeToSplit.GetBalance().Uint64() != 0 {
+		if unstakeInput.SplitRemainder.IsZero() {
```
conorpp, line 191: The transaction input should be considered untrusted.  It would be better if the input provided the list of all necessary staking inputs, and then the trusted transaction builder contained the logic to serialize accordingly.

I.e. the user inputs `args.GetAmount()`, but here it's the remote client effectively telling the builder how much to unstake.  Not as trustworthy.

conorpp: Another alternative is we can keep the client input, but the builder should verify that the resulting amount matches the user input `args.GetAmount()`.

chain/sui/tx.go
```diff
@@ -239,3 +240,85 @@ func (tx Tx) GetMetadata() ([]byte, error) {
...
+	remainingPrincipal := principalDecimal.Sub(amountPrincipal)
+	return xc.NewAmountBlockchainFromUint64(remainingPrincipal.BigInt().Uint64()), true
+}
+
+type UnstakingInput struct {
+	TxInput
+	// Stakes that can be fully unstaked via `request_withdraw_stake`
+	StakesToUnstake []Stake `json:"stakes_to_unstake"`
+	// Stake to split to split for remaining amount
+	StakeToSplit Stake `json:"stakes_to_merge"`
```
conorpp, line 313: fix json key `stake_to_split`

chain/sui/tx.go
```diff
@@ -239,3 +240,85 @@ func (tx Tx) GetMetadata() ([]byte, error) {
...
+}
+
+type Stake struct {
+	Principal xc.AmountBlockchain `json:"principal"`
+	Rewards   xc.AmountBlockchain `json:"rewards"`
+	ObjectId  string              `json:"object_id"`
+	Version   uint64              `json:"version"`
+	Digest    string              `json:"digest"`
+	State     xclient.StakeState  `json:"state"`
+	Validator string              `json:"validator"`
```
conorpp, line 264: This is `repeated`, from `StakingInput.Validator`, is it simplified?  Also don't want to replace user input `StakingArgs.Validator` on the transaction builder.

PR #225, opened 2026-03-12
Implement native canton driver

PawelBis:


chain/canton/address/address_test.go
```diff
@@ -0,0 +1,120 @@
...
+		require.NoError(t, err)
+		require.Equal(t, 68, len(fp), "fingerprint must always be 68 chars")
+		expected := map[byte]string{
+			0:   "1220ea618da83b6c6b2c4557ffa17d722045169f52b8f50f3b31fc867e266de7e53d",
+			1:   "1220974cb80e78f2fea077628a02faa4c57d68a65036eea27fb3463088a1c8527a99",
+			127: "122087fab42073577d1d066f0cc217b347aedf5a73ee5feaa75d5e96538dad977b91",
+			255: "122044e19d94c296e8397d61e759ff1692e5dff8efbcd70d7a9b4033d8b4a259ccd0",
+		}
+		require.Equal(t, expected[seed], fp)
+	}
```
conorpp: little messy, please refactor into full table-based test.

chain/canton/address/address_test.go
```diff
@@ -0,0 +1,120 @@
...
+		expectErr    bool
+		errContains  string
+		expectedName string
+		expectedFP   string
+	}{
+		{
+			name:         "valid - pubkey hex name",
+			addr:         xc.Address("aabbccdd::" + validFP),
+			expectedName: "aabbccdd",
+			expectedFP:   validFP,
```
conorpp, line 76:
```go
expectedFP: "1220" + hex.EncodeToString(make([]byte, 32)), // "1220" + 64 zeros
```

chain/canton/tx_input/tx_input.go
```diff
@@ -0,0 +1,76 @@
...
+)
+
+// TxInput for Canton holds the result of the prepare step in the
+// interactive-submission (external-party signing) flow.
+type TxInput struct {
+	xc.TxInputEnvelope
+	IsExternalTransfer   bool  `json:"is_external_transfer"`
+	LedgerEnd            int64 `json:"ledger_end"`
+	PreparedTransaction  *interactive.PreparedTransaction
+	HashingSchemeVersion interactive.HashingSchemeVersion
```
conorpp:
```go
	PreparedTransaction  *interactive.PreparedTransaction `json:"prepared_transaction"`
	HashingSchemeVersion interactive.HashingSchemeVersion `json:"hashing_scheme_version"`
```

chain/canton/client/client.go
```diff
@@ -0,0 +1,1177 @@
...
+		return fmt.Errorf("failed to unmarshal Canton execute request: %w", err)
+	}
+
+	andWaitReq := &interactive.ExecuteSubmissionAndWaitRequest{
+		PreparedTransaction:  req.PreparedTransaction,
+		PartySignatures:      req.PartySignatures,
+		SubmissionId:         req.SubmissionId,
+		UserId:               req.UserId,
+		HashingSchemeVersion: req.HashingSchemeVersion,
+	}
```
conorpp: Why unmarshal to `ExecuteSubmissionRequest` to then convert to `ExecuteSubmissionAndWaitRequest`?

Wouldn't it be more simple to serialize to `ExecuteSubmissionAndWaitRequest` originally?

chain/canton/client/client.go
```diff
@@ -0,0 +1,1177 @@
...
+	idx := strings.Index(value, "-")
+	if idx <= 0 || idx == len(value)-1 {
+		return 0, "", false
+	}
+	beginExclusive, err := strconv.ParseInt(value[:idx], 10, 64)
+	if err != nil {
+		return 0, "", false
+	}
+	return beginExclusive, value[idx+1:], true
+}
```
conorpp, line 2083: If you refactor the protobuf to be generated locally, then you can refactor all of these helpers to be local methods on the pb types, to support TxInfo method.

E.g.

```
path/to/proto/record.pb.go
path/to/proto/record.go // <-- Add our own helpers in our own files like these
```

chain/canton/client/client.go
```diff
@@ -0,0 +1,1177 @@
...
+func (client *Client) FetchDecimals(ctx context.Context, contract xc.ContractAddress) (int, error) {
+	if client.Asset.GetChain().IsChain(contract) {
+		return int(client.Asset.GetChain().GetDecimals()), nil
+	}
+	return 0, fmt.Errorf("token decimals are not supported for Canton, contract: %s", contract)
+}
+
+func (client *Client) FetchBlock(ctx context.Context, args *xclient.BlockArgs) (*txinfo.BlockWithTransactions, error) {
+	return &txinfo.BlockWithTransactions{}, errors.New("not implemented")
+}
```
conorpp, line 2262: Is there any equivalent grouping of transactions on Canton that we could build on?  Not a big deal if not. 

Perhaps just fetching the latest transaction and reporting a "block" with it?   Return not-support/not-implemented error kind if height is specified.

client/account.go
```diff
@@ -0,0 +1,51 @@
...
+	// GetAccountState returns the current create-account state without mutating it.
+	GetAccountState(ctx context.Context, args *CreateAccountArgs) (*AccountState, error)
+}
+
+type AccountStateEnum string
+
+const (
+	CreateAccountCallRequired AccountStateEnum = "CreateAccountCallRequired"
+	Pending                   AccountStateEnum = "Pending"
+	Created                   AccountStateEnum = "Created"
```
conorpp: Use these values for consistency with other cordialsys apis:

```
inactive // CreateAccountCall required
registering
active
```

chain/canton/client/client.go
```diff
@@ -0,0 +1,2494 @@
...
+	contracts, err := client.LedgerClient.GetActiveContracts(ctx, partyID, ledgerEnd, true)
+	if err != nil {
+		if isPermissionDenied(err) {
+			Logger.WithError(err).Info("get-account-state: party exists but contract visibility is not ready yet")
+			return xclient.AccountInactive, nil
+		}
+		Logger.WithError(err).Error("get-account-state: get active contracts failed")
+		return "", fmt.Errorf("failed to fetch active contracts: %w", err)
+	}
+	testutil.JsonPrint(contracts)
```
shevtsiv: I think this can be removed now

chain/canton/client/client.go
```diff
@@ -0,0 +1,2494 @@
...
+
+	commandID := cantonproto.NewCommandID()
+	synchronizerID, err := client.resolveSynchronizerID(ctx, string(args.GetFrom()), registrySynchronizerID)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to resolve native Amulet transfer synchronizer: %w", err)
+	}
+	actAs := []string{string(args.GetFrom())}
+	readAs := []string{string(args.GetFrom())}
+	prepareReq := cantonproto.NewPrepareRequest(commandID, synchronizerID, actAs, readAs, []*v2.Command{cmd}, disclosedContracts)
+	if client.LedgerClient.Logger != nil {
```
shevtsiv: We check for Logger nil here, but not in the other places, like https://github.com/cordialsys/crosschain/pull/225/changes#diff-988135a7f544611d04864ed1539d22b484be7c6da2535e672d8bee82aac81d49R1883
Might be good to add some small logging function

chain/canton/tx/tx.go
```diff
@@ -0,0 +1,421 @@
...
+		}
+		seen.Set(parentID, true)
+		nodeID = parentID
+	}
+}
+
+// validateCreateNode checks a TransferOffer Create node for matching receiver and amount.
+func validateCreateNode(create *v1.Create, wantReceiver string, decimals int32, wantInstrumentAdmin string, wantInstrumentID string) (xc.AmountBlockchain, bool, error) {
+	tid := create.GetTemplateId()
+	if tid == nil || tid.GetEntityName() != "TransferOffer" {
```
shevtsiv, line 172: Should we also check for `templateID.GetModuleName() == "Splice.Wallet.TransferOffer"` here?

chain/canton/tx/tx.go
```diff
@@ -0,0 +1,421 @@
...
+		return fmt.Errorf("prepared transaction contains no recognized Canton transfer node")
+	}
+
+	totalAmount := xc.NewAmountBlockchainFromUint64(0)
+	for _, transfer := range validatedTransfers {
+		if hasRecognizedTransferAncestor(transfer.nodeID, parentByChildID, validatedTransferIDs) {
+			continue
+		}
+		totalAmount = (&totalAmount).Add(&transfer.amount)
+		if (&totalAmount).Cmp(&wantAmount) > 0 {
```
shevtsiv, line 146: Should we check for a strict match between what was wanted and actual accumulated result and not less/more? I mean after the summing loop

shevtsiv (approved):

shevtsiv (approved):

chain/canton/tx/tx.go
conorpp: would like to just enforce an upper bound for now until we get more testing.

chain/canton/tx/tx.go
conorpp: that is checked via the suffix later on

PR #239, opened 2026-04-24
solana: refactor transaction parsing to be more resuable, improve tok…

conorpp:
…en account resolution

### DIFF
--- a/chain/solana/client/decode.go
+++ b/chain/solana/client/decode.go
@@ -1,0 +1,341 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	xc "github.com/cordialsys/crosschain"
+	soltx "github.com/cordialsys/crosschain/chain/solana/tx"
+	"github.com/cordialsys/crosschain/chain/solana/types"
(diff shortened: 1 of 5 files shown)


chain/solana/client/decode.go
```diff
@@ -0,0 +1,337 @@
...
+			xc.Address(instr.Instruction.GetAuthorityAccount().PublicKey.String()),
+			xc.Address(to),
+			xc.NewAmountBlockchainFromUint64(*instr.Instruction.Amount),
+			contract,
+			txinfo.NewEvent(instr.ID, txinfo.MovementVariantNative),
+		)
+	}
+	if solClient != nil {
+		for _, instr := range decoder.GetCloseTokenAccounts() {
+			lamports, err := solClient.GetMinimumBalanceForRentExemption(ctx, 165, rpc.CommitmentFinalized)
```
shevtsiv: This seems to be static? Let's move out of the loop maybe

conorpp: done

shevtsiv (approved):

PR #270, opened 2026-09-09
tempo: Add fee_contract support for single transfers

shevtsiv:


chain/tempo/builder.go
```diff
@@ -28,11 +30,63 @@ func NewTxBuilder(cfg *xc.ChainBaseConfig) (TxBuilder, error) {
...
+	// Explicitly build a single EVM call: NewTx would select EIP-7702 for a payer.
+	ethTx, err := evmtx.NewSingleTx(args, evmInput, txBuilder.Asset).BuildEthTx()
+	if err != nil {
+		return nil, err
+	}
+	tx, err := newTempoTx(ethTx, feeContract)
+	if err != nil {
+		return nil, err
+	}
+	tx.sender, tx.feePayer = args.GetFrom(), feePayer
```
conorpp: leaky construction.  perhaps just do:

```
newTempoTx(args, ethTx)
```

asset.go
```diff
@@ -346,7 +346,9 @@ func (native NativeAsset) Driver() Driver {
 		return DriverBitcoinLegacy
 	case ZEC, FLUX:
 		return DriverZcash
-	case ARC, AVAX, BNB, CELO, ETH, ETHW, GUSDT, GiwaETH, MATIC, OptETH, ArbETH, BERA, BASE, SeiEVM, MegaETH, MON, HyperEVM, LinETH, XPL, ZeroG, TEMPO, FRAX, RhETH:
+	case TEMPO:
+		return DriverTempo
```
conorpp: please also update the factory/chains/defaults chain YAML files to have matching driver configured.

chain/tempo/builder.go
shevtsiv: Fixed here https://github.com/cordialsys/crosschain/pull/270/commits/39d1706accf5874e862fa808754ba34e7efa4f5a
Thanks

asset.go
shevtsiv: I think they already use tempo https://github.com/cordialsys/crosschain/blob/main/factory/defaults/chains/mainnet.yaml#L1566

https://github.com/cordialsys/crosschain/blob/main/factory/defaults/chains/testnet.yaml#L879

PR #274, opened 2026-09-18
solana: Add v1 transactions support

shevtsiv:


chain/solana/tx_input/version.go
```diff
@@ -0,0 +1,67 @@
...
+const (
+	LamportsPerSignature           uint64 = 5000
+	TransactionVersionLegacy              = "legacy"
+	TransactionVersionV0                  = "v0"
+	TransactionVersionV1                  = "v1"
+	MaxComputeUnitLimit            uint32 = 1_400_000
+	MaxLoadedAccountsDataSizeLimit uint32 = 64 * 1024 * 1024
+)
+
+func (input *TxInput) MessageVersion() (solana.MessageVersion, error) {
```
conorpp: I think we should drop; this client endpoint does not decide the transaction version, the builder does.

chain/solana/tx_input/tx_input.go
```diff
@@ -12,11 +12,18 @@ import (
 // TxInput for Solana
 type TxInput struct {
 	xc.TxInputEnvelope
-	RecentBlockHash     solana.Hash      `json:"recent_block_hash,omitempty"`
-	ToIsATA             bool             `json:"to_is_ata,omitempty"`
-	TokenProgram        solana.PublicKey `json:"token_program"`
-	ShouldCreateATA     bool             `json:"should_create_ata,omitempty"`
-	SourceTokenAccounts []*TokenAccount  `json:"source_token_accounts,omitempty"`
+	// Empty preserves the legacy format used by older inputs.
+	TransactionVersion string `json:"transaction_version,omitempty"`
```
conorpp: I think we should drop; this client endpoint does not decide the transaction version, the builder does.

conorpp: Perhaps rename to `SupportsV1 bool`

chain/solana/tx_input/call.go
```diff
@@ -14,6 +14,39 @@ func init() {
...
+		return input.TxInput.GetFeeLimit()
+	}
+	fee := xc.NewAmountBlockchainFromUint64(0)
+	if input.TransactionConfig.PriorityFee != nil {
+		fee = xc.NewAmountBlockchainFromUint64(*input.TransactionConfig.PriorityFee)
+	}
+	// FetchBaseInput may reserve rent for a nonce account that this prebuilt
+	// call never creates. Only its actual signatures contribute to the base fee.
+	baseFee := xc.NewAmountBlockchainFromUint64(uint64(input.NumRequiredSignatures) * LamportsPerSignature)
+	return fee.Add(&baseFee), ""
```
conorpp: This seems kind of messy to me.  The base `TxInput` already has `.V1Config()` method, but we add a duplicate `.TransactionConfig` in `CallInput`.  Same thing with `NumRequiredSignatures`.

Perhaps the base TxInput could be refactored to store a V1Config redundantly -- that way builders that use v1 can pick it up.  And our CallInput endpoint can populate it & use it more cleanly.

shevtsiv: Cleaned up the redundancy here https://github.com/cordialsys/crosschain/pull/274/commits/9068430c79f26ad5036dd591488732da2c317fe2

chain/solana/tx_input/version.go
shevtsiv: Not dropped, but the builder is the only user now.

conorpp (approved):

